PR #66, opened 2019-09-30
Attempted speed improvements

lodainn:
Here are some attempted speed improvements for the miner inner loop.  I'm not sure that the standard integer will cover the entire solution space.  Should probably be tested a bit.

- Replaced randInt() with simple integer.   Since randInt() has substantial computations, replacing with a simple integer counter should give savings.  The starting integer is random, in a simple attempt to eliminate duplicate work by worker threads (may not be effective all the time).  The resulting submitted nonce is 19 bytes (shorter than the current bigInt nonce).  Hopefully this gives enough 

- Moved some object constructors out of the loop.  The ripemd160.New() was left in because taking it out seemed to cause submission errors.  ...needs more work.  Might try using ripemd160.Reset()

### DIFF
--- a/pow/powWorker.go
+++ b/pow/powWorker.go
@@ -11,6 +11,7 @@ import (
 	"strconv"
 	"sync"
 	"time"
+	"math/rand"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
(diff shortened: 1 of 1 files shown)


mdcoon:
Looks good. We can further optimize if we eliminate the hex conversions but that will take a lot more thought.

PR #86, opened 2019-11-20
Documentation links now directed to read the docs

brendaloya:


brendaloya:
I've added the disclaimer to the read the docs as well.

Issue #80, opened 2019-10-29
Segfaults when malformed config file provided


suspended:
When a malformed config file (missing comma for example) is provided, the binary segfaults without any helpful error message.

A good change would be to detect if the config file is valid on startup before bootstrapping other modules.

Issue #92, opened 2019-12-31
Add validators to config files 


themandalore:
and better documentation on defaults

PR #99, opened 2020-01-11
~3x GPU miner performance, GPU config enabled, basic config validation

charliehorse55:
This increased the speed of the GPU miner on my 970 from 75 to 225 MH/s. 

The format of the config file has changed for GPUs. To enable gpu support, add this in config.json:

```
    "gpuConfig":{
        "<your gpu's name>":{
                "groupSize":64,
                "groups":128,
                "count":256
        }
    },
``` 
When the GPUs are initalized, each device will use a gpuConfig that exactly matches it's name, falling back to a default if it doesn't exist. GPUconfigs matching no devices are silently ignored.  

The exact names of the GPUs in your system are printed at launch. 

You need only add a config entry for each *unique* model of GPU in your system.


charliehorse55:
Fixes #80 #92

PR #122, opened 2020-02-15
Fixed hashrate printout when nonce found

charliehorse55:
Fixes an issue that causes reported hashrate to be much higher than actual when mining on rinkeby.


charliehorse55:
This also now fixes integer overflow in pool.go

Issue #170, opened 2020-08-18
Config file revamp


themandalore:
better defaults (e.g. create default trackers)
Simplify and reduce, creating options, but somehow less

JGcarv:
Can we close this?

PR #179, opened 2020-09-14
WIP config revamp

JGcarv:


JGcarv:
@themandalore I think I finished this PR, which closes #170 

Basically the main changes are that a lot more fields got default values, which are all declared in the config var, on line 90 and the trackers have a different declaration method:

Instead of passing an array of strings, trackers should be nested in the json:
```json
{
    "contractAddress": "0x0Ba45A8b5d5575935B8158a88C631E9F9C95a2e5",
    ...
    "serverPort": 5000,
    "trackers": {
        "indexers": false,
        "disputeChecker": true
    },
    "dbFile": "/tellorDB"
}
```
If a tracker is NOT declared, it takes the default value, otherwise it takes the declared value. In the above example, all defaults are active, except for `indexers` and the optional `disputeChecker` is also active.

Let me know what you think!

PR #189, opened 2020-09-30
Dev

themandalore:


themandalore:
a big one...been running this one on dev.  Will continue to fix and improve, but this is a non-mandatory, mainly cosmetic update to the miner.  If you're miner was getting stuck on challenges (unconfirmed reports), this one seems to help (also word of mouth reports)

PR #204, opened 2020-10-14
Fix the coverage tests

krasi-georgiev:
will add proper e2e tests instead.


krasi-georgiev:
as per the offline discussion, I am not fixing these and will see how it goes.

krasi-georgiev:
tests are now passing, I have sent a request to add the app to the organization so that is posts comments with the coverage results.

themandalore (approved):

PR #209, opened 2020-10-16
WIP initial log structure

JGcarv:
Just a WIP PR to see if I'm in the correct line of implementation, before I spend quite some time replacing the printf statements!

### DIFF
--- a/main.go
+++ b/main.go
@@ -8,7 +8,6 @@ import (
 	"context"
 	"crypto/ecdsa"
 	"fmt"
-	"log"
 	"math/big"
 	"os"
 	"os/signal"
@@ -17,6 +16,8 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
 	cli "github.com/jawher/mow.cli"
 	tellorCommon "github.com/tellor-io/TellorMiner/common"
 	"github.com/tellor-io/TellorMiner/config"
@@ -31,6 +32,15 @@ import (
 
 var ctx context.Context
 
+func setupLogger() log.Logger {
+	lvl := level.AllowError()
+
+	logger := log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr))
+	// logger = level.NewFilter(logger, lvl)
+
+	return log.With(logger, "ts", log.DefaultTimestampUTC, "caller", log.DefaultCaller)
+}
+
 func ErrorHandler(err error, operation string) {
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "%s failed: %s\n", operation, err.Error())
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -265,17 +282,17 @@ func mineCmd(cmd *cli.Cmd) {
 		DB := ctx.Value(tellorCommon.DataProxyKey).(db.DataServerProxy)
 		v, err := DB.Get(db.DisputeStatusKey)
 		if err != nil {
-			fmt.Println("ignoring --- could not get dispute status.  Check if staked")
+			level.Warn(logger).Log("msg", "ignoring --- could not get dispute status.  Check if staked")
 		}
 		status, _ := hexutil.DecodeBig(string(v))
 		if status.Cmp(big.NewInt(1)) != 0 {
-			log.Fatalf("Miner is not able to mine with status %v. Stopping all mining immediately", status)
+			level.Error(logger).Log("err", "Miner is not able to mine with status %v. Stopping all mining immediately", status)
```
krasi-georgiev, line 289: lets be consistent and use `msg` and OTOH in the logger we need to have an even number of argument or use `fmt.Sprintf`

so here you can either have 
`level.Error(logger).Log("msg", fmt.Sprintf("Miner is not able to mine with status %v. Stopping all mining immediately", status))`

or 

`level.Error(logger).Log("msg", "Miner is not able to mine. Stopping all mining immediately","status", status))`
I am a fan of this second version but don't mind either way.

also when you see `log.Fatalf` this means that the should also exit after logging so here you should also add `os.Exit(1)` after the log line.

ops/disputeOps.go
```diff
@@ -205,6 +207,7 @@ func List(ctx context.Context) error {
 
 	instance := ctx.Value(tellorCommon.MasterContractContextKey).(*tellor.TellorMaster)
 
+	// Not sure if we should handle this with logger??
```
krasi-georgiev, line 210: no this one is fine to leave like this as the comment is to just list disputes

ops/miningManager.go
```diff
@@ -40,11 +42,12 @@ type MiningMgr struct {
 }
 
 // CreateMiningManager creates a new manager that mananges mining and data requests.
-func CreateMiningManager(ctx context.Context, exitCh chan os.Signal, submitter tellorCommon.TransactionSubmitter) (*MiningMgr, error) {
+func CreateMiningManager(ctx context.Context, exitCh chan os.Signal, submitter tellorCommon.TransactionSubmitter, logger log.Logger) (*MiningMgr, error) {
```
krasi-georgiev, line 45: looks like you missed my chat comments :smile_cat: 
In golang as standard first argument is context and second is logger.

```suggestion
func CreateMiningManager(ctx context.Context, logger log.Logger, exitCh chan os.Signal, submitter tellorCommon.TransactionSubmitter) (*MiningMgr, error) {
```

ops/miningManager.go
```diff
@@ -40,11 +42,12 @@ type MiningMgr struct {
...
 	cfg := config.GetConfig()
 
 	group, err := pow.SetupMiningGroup(cfg)
 	if err != nil {
+
```
krasi-georgiev, line 50:
```suggestion
```

krasi-georgiev (commented):
Thanks @JGcarv very good start! 
I added some initial comments.

main.go
```diff
@@ -146,11 +156,14 @@ func App() *cli.Cli {
 	configPath := app.StringOpt("config", "configs/config.json", "Path to the primary JSON config file")
 	logPath := app.StringOpt("logConfig", "configs/loggingConfig.json", "Path to a JSON logging config file")
 
+	//Ignoring loging Config for now
```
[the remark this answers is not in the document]
JGcarv: The reason I'm ignoring for now is that the log config allows to set different log filters for each package, which I couldn't quite figure it out how to do yet.

ops/miningManager.go
JGcarv: Ops, I actually missed that. I'll adjust it

main.go
```diff
@@ -213,27 +226,31 @@ func disputeCmd(cmd *cli.Cmd) {
 }
 
 func voteCmd(cmd *cli.Cmd) {
+	logger := setupLogger()
```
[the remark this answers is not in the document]
JGcarv: The this is that for some reason I can't pass any argument to the functions that end with `Cmd` and take the `cmd *cli.Cmd` as a first argument. 

I initially set it up to config on here and pass to all the functions below:
```	app.Command("stake", "staking operations", stakeCmd)
	app.Command("transfer", "send TRB to address", moveCmd(ops.Transfer))
	app.Command("approve", "approve TRB to address", moveCmd(ops.Approve))
	app.Command("balance", "check balance of address", balanceCmd)
	app.Command("dispute", "dispute operations", disputeCmd)
	app.Command("mine", "mine for TRB", mineCmd)
	app.Command("dataserver", "start an independent dataserver", dataserverCmd)
```

But I can't figure it out how to do that

main.go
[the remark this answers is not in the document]
krasi-georgiev: tbh I don't think we even need separate configs for each component, lets just use a single log config and can be even passes as cli args like in Thanos.

main.go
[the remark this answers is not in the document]
krasi-georgiev: you can do it like this.

```
func disputeCmd(logger log.Logger, cmd *cli.Cmd) {
	cmd.Command("vote", "vote on an active dispute", voteCmd(logger))
	...
	..
}

func voteCmd(logger) *cli.Cmd {
	return func(cmd *cli.Cmd) {
		disputeID := EthereumInt{}
		cmd.VarArg("DISPUTE_ID", &disputeID, "dispute id")
		supports := cmd.BoolArg("SUPPORT", false, "do you support the dispute? (true|false)")
		cmd.Action = func() {
			ErrorHandler(ops.Vote(disputeID.Int, *supports, ctx, logger), "vote")
		}
	}
}
```

it is ugly but lets keep it with minimum refactoring and will fix this in follow up prs.

JGcarv:
I Believe I addresses most of this issues in the other PR. This is outdated by now!

PR #211, opened 2020-10-19
Auto genrate the contract bindings.

krasi-georgiev:
- The bindings now use `Tellor.sol` and `TellorGettes.sol`. Generated
into packages called `Tellor` and `TellorGetters` respectively.
- Added makefile target to download latest contracts from tellor core
repo and another makefile target to generate the bindings. Run `make
help` to see all commands
- moved contracts from folder `abi` to folder `contracts` as this is
the default location for `buidler` and will be easier later with the
tests.
- as part of updating the go bindings needed to update few golang
dependencies packages


pkg/pow/solutionHandler.go
```diff
@@ -225,8 +225,9 @@ func (s *SolutionHandler) submit(ctx context.Context, contract tellorCommon.Cont
 
 	txn, err := contract.SubmitSolution(
 		s.currentNonce,
-		s.currentChallenge.RequestID,
-		s.currentValue)
+		[5]*big.Int{s.currentChallenge.RequestID}, // ?????????????
+		[5]*big.Int{s.currentValue},               // ???????????????
```
krasi-georgiev, line 229: It seems that the interface for the `SubmitSolution` has changed from a single value input into an array of 5 values. Is that correct and how do you think we should handle it on the golang side?

pkg/tracker/disputeChecker.go
```diff
@@ -147,8 +147,8 @@ func (c *disputeChecker) Exec(ctx context.Context) error {
 			blockTime = time.Unix(int64(header.Time), 0)
 			blockTimes[l.BlockNumber] = blockTime
 		}
-		reqID := nonceSubmit.RequestId.Uint64()
-		result := CheckValueAtTime(reqID, nonceSubmit.Value, blockTime)
+		reqID := nonceSubmit.RequestId[0].Uint64()                         // ???????????????????????
+		result := CheckValueAtTime(reqID, nonceSubmit.Value[0], blockTime) // ??????????????????
```
krasi-georgiev, line 151: Here this check also needs to be fixed. Shall it just loop over all values and check each one or it should be some median? Is each value about the same currency update or each value is an update for a separate currency?

krasi-georgiev (commented):
@JGcarv , @themandalore I need some input from you on this one regarding the `SubmitSolution` interface change.

pkg/pow/solutionHandler.go
themandalore: yeah, so it's just the array of 5, the single value was v1

pkg/tracker/disputeChecker.go
themandalore: ah, yeah, I don't even know if we've updated disputeChecker for v2

pkg/pow/pool.go
```diff
@@ -136,7 +136,7 @@ func (p *StratumPool) GetWork(input chan *Work) (*Work, bool) {
 					Difficulty: miningNotify.MedianDifficulty,
 					// Difficulty: big.NewInt(10000000),
 					// Difficulty: big.NewInt(6377077812),
-					RequestID: big.NewInt(1),
+					RequestIDs: [5]*big.Int{big.NewInt(1)},
```
krasi-georgiev, line 139: I don't understand this bit.

What is a Stratum pool?

themandalore: I think I was just setting a base value and we set them later on....but not sure

themandalore (approved):

PR #212, opened 2020-10-19
Standardize logging

JGcarv:
I just restarted this because the whole folder structure changed, so I decided it was better to do it now than to rebase everything else later.

I think this time is going a bit better, I managed to create just one instance of `logger` in the main.go and then pass it to functions.

The only thing I'm worried about now, that's why I stopped changing at this point, is that Data Servers uses channels, which believe is some mechanism to run parallel stuff, and I don't know if I need to do some additional work to make the logger function properly. previously it was passed inside an ops structure, and a new instance every time, so I'm not sure if what I'm doing will be fine!


krasi-georgiev:
> The only thing I'm worried about now, that's why I stopped changing at this point, is that Data Servers uses channels, which believe is some mechanism to run parallel stuff, and I don't know if I need to do some additional work to make the logger function properly. previously it was passed inside an ops structure, and a new instance every time, so I'm not sure if what I'm doing will be fine!

yes that is fine there is nothing special with the logger with working with channels, just pass it around as usual and that is it.

cmd/tellor/main.go
```diff
@@ -106,20 +116,22 @@ func AddDBToCtx(remote bool) error {
 	os.RemoveAll(cfg.DBFile)
 	DB, err := db.Open(cfg.DBFile)
 	if err != nil {
-		return err
+		return fmt.Errorf("Error opening DB: %v\n", err)
```
krasi-georgiev, line 119: using errors.Wrap is better here:

```suggestion
		return errors.Wrap(err,"opening the DB")
```

cmd/tellor/main.go
```diff
@@ -233,135 +250,145 @@ func newDisputeCmd(cmd *cli.Cmd) {
...
+
+					var err error
+					ds, err = ops.CreateDataServerOps(ctx, logger, ch)
+					if err != nil {
+						level.Error(logger).Log("msg", "error creating data server", "err", err)
+						os.Exit(1)
+					}
+					// Start and wait for it to be ready.
+					if err := ds.Start(ctx, logger); err != nil {
+						level.Error(logger).Log("msg", "error starting data server", "err", err)
```
krasi-georgiev, line 278: I failed start should also exit here no?

```suggestion
						level.Error(logger).Log("msg", "error starting data server", "err", err)
						os.Exit(1)
```

cmd/tellor/main.go
```diff
@@ -233,135 +250,145 @@ func newDisputeCmd(cmd *cli.Cmd) {
...
-
-			if ds != nil {
-				dsStopped = !ds.Running
-			} else {
-				dsStopped = true
+			// Start miner
+			DB := ctx.Value(tellorCommon.DataProxyKey).(db.DataServerProxy)
+			v, err := DB.Get(db.DisputeStatusKey)
+			if err != nil {
+				level.Warn(logger).Log("info", "could not get dispute status. Check if staked")
```
krasi-georgiev, line 287:
```suggestion
				level.Warn(logger).Log("msg", "could not get dispute status. Check if staked")
```

lets keep the format the same everywhere. See prometheus and thanos for examples.
when you have `level.Info` it automatically adds INFO to the logs. the same for warn and error logs

cmd/tellor/main.go
```diff
@@ -233,135 +250,145 @@ func newDisputeCmd(cmd *cli.Cmd) {
...
+
+				if !dsStopped && !minerStopped && cnt > 60 {
+					level.Warn(logger).Log("msg", "Taking longer than expected to stop operations", "waited", time.Since(start))
+				} else if dsStopped && minerStopped {
+					break
+				}
+				time.Sleep(500 * time.Millisecond)
 			}
-			time.Sleep(500 * time.Millisecond)
+			level.Info(logger).Log("info", "Main shutdown complete")
```
krasi-georgiev, line 335:
```suggestion
			level.Info(logger).Log("msg", "Main shutdown complete")
```

cmd/tellor/main.go
JGcarv: I'll leave this as is right now because this same thing is used all over the code. But if it's worth the time, I can do another PR to adjust those all  in one batch. What do you think?

JGcarv:
I don't understand how to make the linting not fail... It's complaining about not checking for errors in the Log function!

cmd/tellor/main.go
krasi-georgiev: I think it is fine to update gradually, since you are changing the code here anyway why not just use the proper form and the rest can do in follow up PRs.

krasi-georgiev:
> I don't understand how to make the linting not fail... It's complaining about not checking for errors in the Log function!

it is fine to exclude this check. See https://github.com/krasi-georgiev/prometheus/blob/45506841e664665e8f0b1b59f416c91643913a3f/scripts/errcheck_excludes.txt#L10

pkg/tracker/currentVariables_test.go
```diff
@@ -29,8 +30,23 @@ func TestCurrentVariables(t *testing.T) {
 
 	tracker := &CurrentVariablesTracker{}
 
+	ctx := context.WithValue(context.Background(), tellorCommon.ClientContextKey, client)
```
krasi-georgiev, line 33: This should be all done in the `testutil.CreateContext(t)`, no? If it isn't best to add it there.

pkg/tracker/disputeChecker.go
```diff
@@ -147,11 +149,22 @@ func (c *disputeChecker) Exec(ctx context.Context) error {
...
-			if result == nil {
-				disputeLogger.Warn("no value data for reqID %d at %s", reqID, blockTime)
-				continue
+		reqID := nonceSubmit.RequestId.Uint64()
+		result := CheckValueAtTime(reqID, nonceSubmit.Value, blockTime)
+		if result == nil {
+			level.Warn(logger).Log("msg", "no value data", "reqid", reqID, "blockTime", blockTime)
+			continue
+		}
+		if !result.WithinRange {
```
krasi-georgiev, line 158: this seems to be a duplicate of the one below.

pkg/tracker/disputeChecker_test.go
```diff
@@ -43,8 +45,9 @@ func TestDisputeCheckerOutOfRange(t *testing.T) {
 	execEthUsdPsrs(ctx, t, ethUSDPairs)
 	time.Sleep(2 * time.Second)
 	execEthUsdPsrs(ctx, t, ethUSDPairs)
+	ctx = context.WithValue(ctx, tellorCommon.ContractAddress, common.Address{0x0000000000000000000000000000000000000000})
```
krasi-georgiev, line 48: I added this in `testutil.CreateContext(t)` so shouldn't be needed

pkg/tracker/disputeStatus_test.go
```diff
@@ -37,7 +39,7 @@ func TestDisputeStatus(t *testing.T) {
 	tracker := &DisputeTracker{}
 	ctx := context.WithValue(context.Background(), common.ClientContextKey, client)
 	ctx = context.WithValue(ctx, common.DBContextKey, DB)
-	err = tracker.Exec(ctx)
+	err = tracker.Exec(ctx, logger)
```
krasi-georgiev, line 42: for testing where we don't need the logs go kit also offers noop logger

krasi-georgiev:
at the end can also remove the tellor log util package  `pkg/util/logger.go`

pkg/tracker/disputeChecker_test.go
JGcarv: Ok! I'll remove it.

pkg/ops/disputeOps.go
```diff
@@ -171,7 +172,7 @@ func getNonceSubmissions(ctx context.Context, valueBlock *big.Int, dispute *tell
 	return timedValues, nil
 }
 
-func List(ctx context.Context) error {
+func List(ctx context.Context, logger log.Logger) error {
```
krasi-georgiev, line 175:
```suggestion
func List(ctx context.Context) error {
```

as discussed it looks like we don't need a logger here

pkg/tracker/balance.go
```diff
@@ -43,10 +43,12 @@ func (b *BalanceTracker) Exec(ctx context.Context) error {
...
 
 	if err != nil {
-		fmt.Println("balance Error, balance.go")
+		level.Error(logger).Log("msg", "error getting balance", "err", err)
 		return err
 	}
 	enc := hexutil.EncodeBig(balance)
-	log.Printf("Balance: %v", enc)
+
+	// log.Printf("Balance: %v", enc)
```
krasi-georgiev, line 51:
```suggestion
```

pkg/testutil/testutil.go
```diff
@@ -55,21 +57,30 @@ func CreateContext(t *testing.T) (context.Context, *config.Config, func()) {
...
 	if err != nil {
-		log.Fatal(err)
+		t.Fatalf("Problem creating proxy: %v\n", err)
 	}
 	ctx = context.WithValue(ctx, common.DataProxyKey, proxy)
 
 	return ctx, cfg, cleanup
 }
+
+func SetupLogger() log.Logger {
```
krasi-georgiev, line 79: instead of duplicating the func why not put it in `pkg/util/logger.go` and then can use for the tests and in the main entry point.

pkg/tracker/disputeStatus.go
```diff
@@ -78,13 +79,14 @@ func (b *DisputeTracker) Exec(ctx context.Context) error {
 		//fmt.Println("Getting staker info for address", addr)
 		status, _, err := instance.GetStakerInfo(nil, address)
 		if err != nil {
-			fmt.Printf("Could not get staker dispute status for miner address %s: %v\n", addr, err)
+			level.Error(logger).Log("msg", "Could not get staker dispute status for miner", "address", addr, "err", err)
 		}
 		fmt.Printf("Whitelisted Miner %s Dispute Status: %v\n", addr, status)
```
krasi-georgiev, line 84:
```suggestion
```

pkg/tracker/runner.go
```diff
@@ -31,15 +31,17 @@ func NewRunner(client rpc.ETHClient, db db.DB) (*Runner, error) {
...
+func (r *Runner) Start(ctx context.Context, logger log.Logger, exitCh chan int) error {
 	cfg := config.GetConfig()
 	trackerNames := cfg.Trackers
 	var trackers []Tracker
 	for name, activated := range trackerNames {
 		if activated {
 			t, err := createTracker(name)
 			if err != nil {
-				runnerLog.Error("Problem creating tracker: %s\n", err.Error())
+				level.Warn(logger).Log("msg", "Problem creating tracker", "name", name, "err", err)
```
krasi-georgiev, line 42: I don't think it makes sense to continue if one of the trackers fails to create.
Not everyone looks at the logs to I think this should propagate the error and the caller should decide what to to.

pkg/tracker/runner.go
```diff
@@ -53,21 +55,22 @@ func (r *Runner) Start(ctx context.Context, exitCh chan int) error {
...
+	level.Info(logger).Log("msg", fmt.Sprintf("Created %d trackers", len(trackers)))
 
 	var err error
 	masterInstance := ctx.Value(tellorCommon.ContractsTellorContextKey)
 	if masterInstance == nil {
 		contractAddress := common.HexToAddress(cfg.ContractAddress)
 		masterInstance, err = tellor.NewTellor(contractAddress, r.client)
 		if err != nil {
-			runnerLog.Error("Problem creating tellor master instance: %v\n", err)
+			//I believe the dataSerever will handle the os.Exit(1)
```
krasi-georgiev, line 66: yeah correct here I don't we should log, but just return the message as an error.

pkg/tracker/runner.go
```diff
@@ -102,9 +105,10 @@ func (r *Runner) Start(ctx context.Context, exitCh chan int) error {
 					//runnerLog.Info("Running trackers...")
```
krasi-georgiev, line 105: all of these commented logs should be debug logs, but lets leave for another PR.

pkg/tracker/runner.go
```diff
@@ -102,9 +105,10 @@ func (r *Runner) Start(ctx context.Context, exitCh chan int) error {
...
 					go func(count int) {
 						idx := count % len(trackers)
-						err := trackers[idx].Exec(ctx)
+						err := trackers[idx].Exec(ctx, logger)
 						if err != nil {
-							runnerLog.Error("Problem in tracker %s: %v\n", trackers[idx].String(), err)
+							level.Warn(logger).Log("msg", "problem in traker", "tracker", trackers[idx].String(), "err", err)
+							//runnerLog.Error("Problem in tracker %s: %v\n", trackers[idx].String(), err)
```
krasi-georgiev, line 111:
```suggestion
```

pkg/tracker/timeOut.go
```diff
@@ -46,13 +48,16 @@ func (b *TimeOutTracker) Exec(ctx context.Context) error {
...
+		level.Error(logger).Log("msg", "instance Error, disputeStatus", "err", err)
 		return err
 	}
 	address := "000000000000000000000000" + _fromAddress[2:]
 	decoded, err := hex.DecodeString(address)
 	if err != nil {
-		log.Fatal(err)
+		level.Error(logger).Log("msg", "error decoding address", "err", err)
+		os.Exit(1)
+		//Does log.Fatal terminates the execution automatically?
```
krasi-georgiev, line 59: yes that is why in practise here we shouldn't exit but just pass propagate the error to the caller.

krasi-georgiev (commented):
you can now remove the `logConfig_test.go` and also the `config/loggingConfig.json`

There are few other small things and it is ready for merging.

cmd/tellor/main.go
JGcarv: I think it'll be easier to do it in a new one This way we won't add a lot of new changes to this already big Pr!

pkg/dataServer/dataServer.go
```diff
@@ -31,18 +33,21 @@ type DataServer struct {
...
-func CreateServer(ctx context.Context) (*DataServer, error) {
+func CreateServer(ctx context.Context, logger log.Logger) (*DataServer, error) {
 	cfg := config.GetConfig()
 
 	DB := ctx.Value(common.DBContextKey).(db.DB)
 	client := ctx.Value(common.ClientContextKey).(rpc.ETHClient)
 	run, err := tracker.NewRunner(client, DB)
 	if err != nil {
-		log.Fatal(err)
+		level.Error(logger).Log("msg", "erro rcreating tracker runner instance", "err", err)
```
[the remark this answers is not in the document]
JGcarv: Question, in this case, the function expects the return `(*DataServer, error)`. Should we return `(nil, errorsNew())`?

pkg/ops/disputeOps.go
JGcarv: The issue here is that the function `simpleCmd` in main expect as a parameter a `func(context.Context, log.Logger) error`, if I change this I'll have to add another function in main to support this. I personally think leaving this here is less impactful than adding another function there. What do you think?

pkg/tracker/currentVariables.go
```diff
@@ -49,43 +47,43 @@ func (b *CurrentVariablesTracker) Exec(ctx context.Context) error {
...
 		bitSetVar = []byte{1}
 	}
-	currentVarsLog.Info("Retrieved variables. challengeHash: %x", currentChallenge)
+	level.Info(logger).Log("msg", "Retrieved variables", "challengeHash", currentChallenge)
 
 	err = DB.Put(db.CurrentChallengeKey, currentChallenge[:])
 	if err != nil {
-		fmt.Println("Current Variables Put Error")
+		level.Error(logger).Log("msg", "Current Variables Put Error", "var", "currentChallenge", "err", err)
 		return err
```
krasi-georgiev, line 62: I think we should simplify this and the following lines and just return an error here and then the caller can print the log.
```suggestion
		return errors.Wrap(err,"currentChallenge variables put")
```

pkg/tracker/disputeChecker.go
```diff
@@ -91,7 +90,7 @@ func CheckValueAtTime(reqID uint64, val *big.Int, at time.Time) *ValueCheckResul
 	}
 }
 
-func (c *disputeChecker) Exec(ctx context.Context) error {
+func (c *disputeChecker) Exec(ctx context.Context, logger log.Logger) error {
```
krasi-georgiev, line 93: Let's add a constructor here so that we can have the component name. the same in all other places for each component.

This
```
func (c *disputeChecker) Exec(ctx context.Context, logger log.Logger){
	logger.
}
```
should change into
```
func NewDisputeChecker(logger log.Logger) *DisputeChecker{
	return &disputeChecker {
		logger:  log.With(logger, "component", "dispute checker")
	}
}

func (c *DisputeChecker) Exec(ctx context.Context){
	c.logger
}
``

File: pkg/tracker/disputeChecker_test.go [hunk 21]
```diff
@@ -65,8 +71,9 @@ func TestDisputeCheckerOutOfRange(t *testing.T) {
 }
 
 func execEthUsdPsrs(ctx context.Context, t *testing.T, psrs []*IndexTracker) {
+	logger := util.SetupLogger("debug")
 	for _, psr := range psrs {
-		err := psr.Exec(ctx)
+		err := psr.Exec(ctx, logger)
```
krasi-georgiev, line 76: Yeah here we also need to pass the logger to the constructor(so that we can have the component name in the logs) and not the actual struct method.

pkg/tracker/gas.go
```diff
@@ -36,7 +37,7 @@ func (b *GasTracker) String() string {
 	return "GasTracker"
 }
 
-func (b *GasTracker) Exec(ctx context.Context) error {
+func (b *GasTracker) Exec(ctx context.Context, logger log.Logger) error {
```
krasi-georgiev, line 40: Here we also need a constructor which takes a logger so that can have the component name in the log lines.

The same for all other places.

pkg/tracker/newCurrentVariables.go
```diff
@@ -53,7 +55,7 @@ func (b *NewCurrentVariablesTracker) Exec(ctx context.Context) error {
 	instanceGetter := ctx.Value(tellorCommon.ContractsGetterContextKey).(*getter.TellorGetters)
 	myStatus, err := instanceGetter.DidMine(nil, returnNewVariables.Challenge, fromAddress)
 	if err != nil {
-		fmt.Println("My Status Retrieval Error")
+		level.Error(logger).Log("msg", "My Status Retrieval Error", "err", err)
```
krasi-georgiev, line 58: For all errors, there is a golang practice to use all lower case and I like to do the same for the logging. Also, I don't put words like `error` in the message as it gets repeated twice in the final output.
```suggestion
		level.Error(logger).Log("msg", "my status retrieval", "err", err)
```

pkg/ops/dataServerOps.go
```diff
@@ -19,31 +20,31 @@ type DataServerOps struct {
...
-func CreateDataServerOps(ctx context.Context, exitCh chan os.Signal) (*DataServerOps, error) {
-	ds, err := dataServer.CreateServer(ctx)
+func CreateDataServerOps(ctx context.Context, logger log.Logger, exitCh chan os.Signal) (*DataServerOps, error) {
+	ds, err := dataServer.CreateServer(ctx, log.With(logger, "component", "ops"))
 	if err != nil {
 		return nil, err
 	}
 	done := make(chan int)
-	ops := &DataServerOps{exitCh: exitCh, server: ds, log: util.NewLogger("ops", "DataServerOps"), done: done, Running: false}
+	ops := &DataServerOps{exitCh: exitCh, server: ds, log: logger, done: done, Running: false}
```
krasi-georgiev, line 34: Since this is the constructor we want to add the new logger fields here.
```suggestion
	ops := &DataServerOps{exitCh: exitCh, server: ds, log: log.With(logger, "component", "ops"), done: done, Running: false}
```

ping me if I didn't manage to explain properly.

pkg/ops/dataServerOps.go
```diff
@@ -19,31 +20,31 @@ type DataServerOps struct {
...
 
 	return ops, nil
 }
 
 // Start the data server.
 func (ops *DataServerOps) Start(ctx context.Context) error {
-	if err := ops.server.Start(ctx, ops.done); err != nil {
+	if err := ops.server.Start(ctx, ops.log, ops.done); err != nil {
```
krasi-georgiev, line 41: the logger was already added to the `ops.server` with the `dataServer.CreateServer` constructor.

pkg/tracker/balance_test.go
```diff
@@ -14,10 +14,11 @@ import (
 	"github.com/tellor-io/TellorMiner/pkg/common"
 	"github.com/tellor-io/TellorMiner/pkg/db"
 	"github.com/tellor-io/TellorMiner/pkg/rpc"
+	"github.com/tellor-io/TellorMiner/pkg/util"
 )
 
 func TestStringId(t *testing.T) {
-	tracker := &BalanceTracker{}
+	tracker := &BalanceTracker{logger: util.SetupLogger("debug")}
```
krasi-georgiev, line 21: To be consistent lets use the constructor: `NewBalanceTracker`

pkg/tracker/currentVariables.go
```diff
@@ -5,29 +5,34 @@ package tracker
...
 
 type CurrentVariablesTracker struct {
+	logger log.Logger
 }
 
 func (b *CurrentVariablesTracker) String() string {
 	return CurrentVariablesTrackerName
 }
 
+func New_CurrentVariablesTracker(logger log.Logger) *CurrentVariablesTracker {
```
krasi-georgiev, line 30:
```suggestion
func NewCurrentVariablesTracker(logger log.Logger) *CurrentVariablesTracker {
```

pkg/tracker/currentVariables.go
```diff
@@ -49,44 +54,37 @@ func (b *CurrentVariablesTracker) Exec(ctx context.Context) error {
...
-		fmt.Println("My Status Retrieval Error")
-		return err
+		return errors.Wrap(err, "Status Retrieval")
 	}
 	bitSetVar := []byte{0}
 	if myStatus {
 		bitSetVar = []byte{1}
 	}
-	currentVarsLog.Info("Retrieved variables. challengeHash: %x", currentChallenge)
+	level.Info(b.logger).Log("msg", "Retrieved variables", "challengeHash", currentChallenge)
```
krasi-georgiev, line 63: lets be consistent with all lower case for errors and logs.
```suggestion
	level.Info(b.logger).Log("msg", "retrieved variables", "challengeHash", currentChallenge)
```

pkg/tracker/disputeChecker_test.go
```diff
@@ -35,16 +38,17 @@ func TestDisputeCheckerOutOfRange(t *testing.T) {
...
+	disputeChecker := &disputeChecker{lastCheckedBlock: 500, logger: util.SetupLogger("debug")}
 	if _, err := BuildIndexTrackers(); err != nil {
 		t.Fatal(err)
 	}
 	ethUSDPairs := indexes["ETH/USD"]
 	execEthUsdPsrs(ctx, t, ethUSDPairs)
 	time.Sleep(2 * time.Second)
 	execEthUsdPsrs(ctx, t, ethUSDPairs)
-
+	ctx = context.WithValue(ctx, tellorCommon.ContractAddress, common.Address{0x0000000000000000000000000000000000000000})
```
krasi-georgiev, line 49: Why do we need this? It should be in the CreateContext func

pkg/tracker/disputeStatus.go
```diff
@@ -49,21 +57,21 @@ func (b *DisputeTracker) Exec(ctx context.Context) error {
 
 	instance, err := getter.NewTellorGetters(contractAddress, client)
 	if err != nil {
-		fmt.Println("instance Error, disputeStatus")
+		level.Error(b.logger).Log("msg", "Error getting master instance", "err", err)
```
krasi-georgiev, line 60: the `error` word is already added by the logger so no need to have it twice, The same in few other places. Also words like `problem`, `Could not` are also not necessary and when reading the logs are just noise. The logger itself formats errors in a very good an distinct format.
```suggestion
		level.Error(b.logger).Log("msg", "getting master instance", "err", err)
```

pkg/tracker/disputeStatus.go
```diff
@@ -49,21 +57,21 @@ func (b *DisputeTracker) Exec(ctx context.Context) error {
...
 		return err
 	}
 
 	status, _, err := instance.GetStakerInfo(nil, fromAddress)
 
 	if err != nil {
-		fmt.Println("instance Error, disputeStatus")
+		level.Error(b.logger).Log("msg", "Error getting staker info ", "err", err)
```
krasi-georgiev, line 67:
```suggestion
		level.Error(b.logger).Log("msg", "getting staker info ", "err", err)
```

pkg/tracker/disputeStatus_test.go
```diff
@@ -14,10 +14,11 @@ import (
 	"github.com/tellor-io/TellorMiner/pkg/common"
 	"github.com/tellor-io/TellorMiner/pkg/db"
 	"github.com/tellor-io/TellorMiner/pkg/rpc"
+	"github.com/tellor-io/TellorMiner/pkg/util"
 )
 
 func TestDisputeString(t *testing.T) {
-	tracker := &DisputeTracker{}
+	tracker := &DisputeTracker{logger: util.SetupLogger("debug")}
```
krasi-georgiev, line 21: When possible and to be consistent lets use the constructor func to be consistent.

pkg/tracker/runner.go
```diff
@@ -31,16 +31,15 @@ func NewRunner(client rpc.ETHClient, db db.DB) (*Runner, error) {
 }
 
 // Start will kick off the runner until the given exit channel selects.
-func (r *Runner) Start(ctx context.Context, exitCh chan int) error {
+func (r *Runner) Start(ctx context.Context, logger log.Logger, exitCh chan int) error {
```
krasi-georgiev, line 34: We should add a constructor here as well and add the logger to the instructor like in all the other changes.
```suggestion
func (r *Runner) Start(ctx context.Context, exitCh chan int) error {
```

pkg/tracker/timeOut.go
krasi-georgiev: this still stands, 
would you mind letting me click the resolve button? 
This way it is easier for me to not miss any of the suggestions and changes. 
If you think that some suggestion is incorrect please comment back and will discuss.

pkg/tracker/timeOut_test.go
```diff
@@ -14,10 +14,11 @@ import (
 	"github.com/tellor-io/TellorMiner/pkg/common"
 	"github.com/tellor-io/TellorMiner/pkg/db"
 	"github.com/tellor-io/TellorMiner/pkg/rpc"
+	"github.com/tellor-io/TellorMiner/pkg/util"
 )
 
 func TestTimeOutString(t *testing.T) {
-	tracker := &TimeOutTracker{}
+	tracker := &TimeOutTracker{logger: util.SetupLogger("debug")}
```
krasi-georgiev, line 21: use the constructor?

pkg/tracker/currentVariables.go
JGcarv: This one is a hack because we already have another tracker call `NewCurrentVariables`, so this is the least awkward way of solving that. But I'm open to suggestions!

pkg/tracker/timeOut.go
JGcarv: this one I don't understand what I'm supposed to change? Should I add  `os.Exit(1)` on all thE errors on this file or just return `errors.Wrapf(...)`?

krasi-georgiev: yes return `errors.Wrap`

Exit should happen only in the main package. All other errors should be either logged(when not that important) or returned to the caller

pkg/tracker/currentVariables.go
krasi-georgiev: aah yeah this the v1 code. The `pkg/tracker/currentVariables.go` file can be removed and this will resolve the conflict. If you prefer can do it in a parallel PR or can do it even in this one.

cmd/tellor/main.go
```diff
@@ -44,17 +46,19 @@ func buildContext() error {
 		// Create an rpc client
 		client, err := rpc.NewClient(cfg.NodeURL)
 		if err != nil {
-			log.Fatal(err)
+			return fmt.Errorf("Couldn't create client instance: %s", err)
```
krasi-georgiev, line 49: here and few other places, lets use errors.Wrap, all lower case, and no need to use words like - error, couldn't , problem etc. these are obvious from the error log.
```suggestion
			return errors.Wrap(err,"create client instance")
```

cmd/tellor/main.go
```diff
@@ -218,139 +227,149 @@ func newDisputeCmd(cmd *cli.Cmd) {
...
+			ch := make(chan os.Signal)
+			var err error
+			ds, err = ops.CreateDataServerOps(ctx, logger, ch)
+			if err != nil {
+				level.Error(logger).Log("msg", "error creating data server", "err", err)
+				os.Exit(1)
+			}
+			// Start and wait for it to be ready
+			if err := ds.Start(ctx, logger); err != nil {
+				//Should we do this here or pass it down to ExitOnError func for consistency?
```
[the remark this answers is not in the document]
krasi-georgiev: this one still stands.

pkg/tracker/disputeStatus.go
```diff
@@ -78,15 +86,15 @@ func (b *DisputeTracker) Exec(ctx context.Context) error {
 		//fmt.Println("Getting staker info for address", addr)
 		status, _, err := instance.GetStakerInfo(nil, address)
 		if err != nil {
-			fmt.Printf("Could not get staker dispute status for miner address %s: %v\n", addr, err)
+			level.Error(b.logger).Log("msg", "getting staker dispute status for miner", "address", addr, "err", err)
 		}
 		fmt.Printf("Whitelisted Miner %s Dispute Status: %v\n", addr, status)
```
krasi-georgiev, line 91:
```suggestion
```

pkg/tracker/factory.go
```diff
@@ -3,45 +3,49 @@
 
 package tracker
 
-import "fmt"
+import (
+	"fmt"
+
```
krasi-georgiev, line 8:
```suggestion
```

pkg/tracker/gas.go
```diff
@@ -55,27 +64,27 @@ func (b *GasTracker) Exec(ctx context.Context) error {
 		if err != nil {
 			gasPrice, err = client.SuggestGasPrice(context.Background())
 			if err != nil {
-				log.Println("couldn't get suggested gas price", err)
+				level.Warn(b.logger).Log("msg", "couldn't get suggested gas price", "err", err)
```
krasi-georgiev, line 67: hm, I think for all trackers we want to return the error to the caller and the caller can decide whether to log it or pass it up the stack. This way we don't even need a logger inside the trackers.
```suggestion
				return errors.Wrap(err, "get suggested gas price")
```

pkg/tracker/balance.go
```diff
@@ -43,10 +50,11 @@ func (b *BalanceTracker) Exec(ctx context.Context) error {
 	balance, err := client.BalanceAt(ctx, fromAddress, nil)
 
 	if err != nil {
-		fmt.Println("balance Error, balance.go")
+		level.Error(b.logger).Log("msg", "getting balance", "err", err)
 		return err
```
krasi-georgiev, line 54: In places where the functions returns the error and stop executing, don't need to log, just pass the error to the caller and let the caller decide whether to log it or pass it up the stack. This way we can even avoid the need to have a logger in functions that just need to return an error. There are few places like this.

Actually in the trackers I think only the gas tracker needs to log as when it fails to get the price from the gasStation website it can get it from the eth client. In these cases where the execution after an error can continue it is ok to print a log otherwise in functions that stop the execution on every error need to just return the error to the caller and no need to print a log.

This way you can see that `CurrentVariablesTracker`, `TributeTracker` don't even need a logger because they just stop the execution on an error and return the error to the caller and the caller can decide whether to print a log or to pass the error up the stack.

You can leave the constructors though, I am sure we will need a constructor at some point anyway.

```suggestion
		return errors.Wrap(err, "getting balance")
```

pkg/tracker/disputeStatus.go
```diff
@@ -78,15 +86,15 @@ func (b *DisputeTracker) Exec(ctx context.Context) error {
 		//fmt.Println("Getting staker info for address", addr)
```
krasi-georgiev, line 86:
```suggestion
```

pkg/tracker/timeOut.go
```diff
@@ -81,14 +88,14 @@ func (b *TimeOutTracker) Exec(ctx context.Context) error {
...
-			log.Fatal(err)
+			return errors.Wrapf(err, "decoding address")
 		}
 		hash := solsha3.SoliditySHA3(decoded)
 		var data [32]byte
 		copy(data[:], hash)
 		status, err := instance.GetUintVar(nil, data)
 		if err != nil {
-			fmt.Printf("Could not get staker timeOut status for miner address %s: %v\n", addr, err)
+			level.Error(b.logger).Log("msg", "getting staker timeOut status for miner", "address", addr, "err", err)
```
krasi-georgiev, line 98: I am not sure about this one. Why would it continue the execution when there is an error?
This means that it will save `0` in the DB. @themandalore can probably give some context here.

pkg/tracker/tributeBalance.go
```diff
@@ -69,9 +77,9 @@ func (b *TributeTracker) Exec(ctx context.Context) error {
 	}
 
 	//numTributes, _ := balanceInTributes.Float64()
-	log.Printf("Tribute Balance: %v (%v tributes)\n", balance, balanceInTributes)
+	fmt.Printf("Tribute Balance: %v (%v tributes)\n", balance, balanceInTributes)
```
krasi-georgiev, line 80: convert to a log

pkg/tracker/factory.go
JGcarv: The extra line is put automatically when I save by the linter... It isn't like this for you?

pkg/tracker/gas.go
JGcarv: Won't we need the logger in case of Info and Debug messages anyway? 

If not, I don't mind removing the logger from trackers, but can we do it in a fresh PR?

pkg/tracker/currentVariables.go
JGcarv: I adjusted the naming but left the tests commented. I'll adjust those in a next PR

pkg/tracker/gas.go
krasi-georgiev: Yes you are right I just double-checked and some are printing some debug logs.

krasi-georgiev (commented):
There are still few places with `fmt.Errorf`, but I opened an issue to do a global search and replace for those https://github.com/tellor-io/TellorMiner/issues/236

krasi-georgiev (approved):

PR #224, opened 2020-10-25
remove the check for manual data as this is done in the indexers

krasi-georgiev:
Some leftover from when I removed the duplicate manual data handling.

### DIFF
--- a/pkg/pow/solutionHandler.go
+++ b/pkg/pow/solutionHandler.go
@@ -56,7 +56,6 @@ func (s *SolutionHandler) Submit(ctx context.Context, result *Result) bool {
 	nonce := result.Nonce
 	s.currentChallenge = challenge
 	s.currentNonce = nonce
-	manualVal := int64(0)
 
 	s.log.Info("Getting pending txn and value from data server...")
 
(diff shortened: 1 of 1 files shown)


krasi-georgiev:
btw can you explain what is the logic here for checking this?
https://github.com/tellor-io/TellorMiner/pull/224/files#diff-6fffd1e0a4f954a84cf8aa7e9e739a248c35e0bc775065e4ba698a09980ba7d2R97

themandalore (approved):

themandalore:
> btw can you explain what is the logic here for checking this?
> 
> ```
> if challenge.RequestIDs[i].Uint64() > 53 {
> 				s.log.Error("Problem decoding price value prior to submitting solution: %v\n", err)					s.log.Error("Problem decoding price value prior to submitting solution: %v\n", err)
> ```
> 
> https://github.com/tellor-io/TellorMiner/pull/224/files#diff-6fffd1e0a4f954a84cf8aa7e9e739a248c35e0bc775065e4ba698a09980ba7d2R97

It's in case a requestID comes through that you don't support.  Instead of 53, it should be the max psr...just a quick check

Issue #186, opened 2020-09-29
Tellor v2.5 and profitability calculation


themandalore:
Check the current reward / tips * TRB price and verify that it is greater than the current gas costs for mining the block
caveat - figure out if it's the fifth block
Add config option to allow users to manually set profitability threshold (e.g. 1.1, .5, etc)

mikeghen:
+1 on this issue

PR #228, opened 2020-10-26
Added profitability calcs and added profitThreshold in config

krasi-georgiev:
Each transaction cost is saved in the db and at the next transaction for
the same slot if the profit is below profitThreshold the transaction is
not executed.

Profit calculations are the same as in the tellor contract.

also moved the GWEI and WEI constants in the common package so that any
package can use them without circular dependencies.

closes: https://github.com/tellor-io/TellorMiner/issues/186


pkg/pow/cpuMiner.go
```diff
@@ -52,19 +50,3 @@ func (c *CpuMiner) CheckRange(hash *HashSettings, start uint64, n uint64) (strin
 	}
 	return "", n, nil
 }
-
-func hashFn(data []byte, result *big.Int) error {
```
krasi-georgiev, line 56: moved to contractRPC

pkg/pow/pool.go
```diff
@@ -159,7 +162,7 @@ func (p *StratumPool) GetWork(input chan *Work) (*Work, bool) {
 	return nil, false
 }
 
-func (p *StratumPool) Submit(ctx context.Context, result *Result) bool {
+func (p *StratumPool) Submit(ctx context.Context, result *Result) (*types.Transaction, error) {
```
krasi-georgiev, line 165: not sure what this method actually does since it doesn' submit anything to the tellor contract.

pkg/pow/solutionHandler.go
```diff
@@ -51,73 +52,62 @@ func CreateSolutionHandler(cfg *config.Config, submitter tellorCommon.Transactio
 	}
 }
 
-func (s *SolutionHandler) Submit(ctx context.Context, result *Result) bool {
+func (s *SolutionHandler) Submit(ctx context.Context, result *Result) (*types.Transaction, error) {
```
krasi-georgiev, line 55: need to return the transaction data to use the gas cost when saving the tx cost in the db.

pkg/rpc/client.go
```diff
@@ -31,6 +31,9 @@ type ETHClient interface {
 	// between contract internal errors and the local chain being out of sync.
 	CodeAt(ctx context.Context, contract common.Address, blockNumber *big.Int) ([]byte, error)
 
+	// TransactionReceipt implements the geth backend DeployBackend interface.
+	TransactionReceipt(ctx context.Context, txHash common.Hash) (*types.Receipt, error)
```
krasi-georgiev, line 35: This is added so that the eth client monitors the completion of a transaction. The transaction gas used value is then saved in the db.

pkg/rpc/contractRPC.go
```diff
@@ -6,26 +6,24 @@ package rpc
...
 	tellorCommon "github.com/tellor-io/TellorMiner/pkg/common"
 	"github.com/tellor-io/TellorMiner/pkg/config"
 	"github.com/tellor-io/TellorMiner/pkg/contracts/getter"
 	"github.com/tellor-io/TellorMiner/pkg/contracts/tellor"
-
 	"github.com/tellor-io/TellorMiner/pkg/db"
-)
-
-var (
-	GWEI = int64(1000000000)
```
krasi-georgiev, line 28: moved to the common package so that other packages can use it. Without moving it caused circular dependency.

pkg/rpc/contractRPC.go
```diff
@@ -49,67 +47,63 @@ func (c contractWrapper) DidMine(challenge [32]byte) (bool, error) {
 	return c.TellorGetters.DidMine(nil, challenge, c.fromAddress)
 }
 
-func PrepareContractTxn(ctx context.Context, proxy db.DataServerProxy, ctxName string, callback tellorCommon.TransactionGeneratorFN) error {
+func SubmitContractTxn(ctx context.Context, proxy db.DataServerProxy, ctxName string, callback tellorCommon.TransactionGeneratorFN) (*types.Transaction, error) {
```
krasi-georgiev, line 50: renamed it becuase it doesn't just prepare but also submits the tx.

pkg/tracker/ampl.go
```diff
@@ -30,6 +30,10 @@ func (a Ampl) ValueAt(vals map[string]apiOracle.PriceInfo, at time.Time) float64
 	return VolumeWeightedAvg(valSlice).Price * a.granularity
 }
 
+func (s *Ampl) Granularity() int64 {
```
krasi-georgiev, line 33: granularity is exposed as it is used when calculating the TRB price.

pkg/tracker/psrs.go
```diff
@@ -15,6 +15,8 @@ import (
 
 var switchTime, _ = time.Parse(time.RFC3339, "2020-06-26T00:00:00+00:00")
 
+const RequestID_TRB_ETH int = 43
```
krasi-georgiev, line 18: this is exposed as a const so that can later get the granularity for the TRB to calculate the TRB price.

krasi-georgiev:
There seems to be an issue with the submission as some transactions fail with `Fail with error 'Request ID is wrong'`  so please hold it for now.

https://rinkeby.etherscan.io/tx/0xc73bb449c9a4968a8fd4d4e379981db31dcf58645b3b13612eaf4fa74e23a34d

pkg/ops/miningManager.go
```diff
@@ -121,26 +169,167 @@ func (mgr *MiningMgr) Start(ctx context.Context) {
...
+				}
 				if result == nil {
 					mgr.Running = false
 					return
 				}
 				mgr.solution = result
-				goodSubmit := mgr.solHandler.Submit(ctx, mgr.solution)
-				if goodSubmit {
-					fmt.Println("solution submitted") //mgr.solution = nil
+				tx, err := mgr.solHandler.Submit(ctx, mgr.solution)
```
themandalore, line 192: move profit check here (below setting mgr.solution so it doesn't have to resolve the solution)

pkg/ops/miningManager.go
```diff
@@ -121,26 +169,167 @@ func (mgr *MiningMgr) Start(ctx context.Context) {
...
+			mgr.log.Error("waiting for transaction completion for  calculating transaction cost err:%v", err)
+		}
+		if receipt.Status != 1 {
+			mgr.log.Error("skiping unsuccessful transaction for calculating transaction cost err:%v", err)
+		}
+		mgr.log.Debug("submit transaction receipt", "txHash:%v", receipt.TxHash)
+		gasUsed := big.NewInt(int64(receipt.GasUsed))
+		txCost := gasUsed.Mul(gasUsed, tx.GasPrice())
+
+		slotNum, err := mgr.contractGetter.GetUintVar(nil, rpc.Keccak256([]byte("slotProgress")))
```
themandalore, line 293: add issue for event listener for nonceSubmitted events to determine slot Progress and whether it was yours.

pkg/config/config.go
```diff
@@ -85,7 +85,7 @@ type Config struct {
 	ConfigFolder                 string                `json:"configFolder"`
 	DisputeTimeDelta             Duration              `json:"disputeTimeDelta"` // Ignore data further than this away from the value we are checking.
 	DisputeThreshold             float64               `json:"disputeThreshold"` // Maximum allowed relative difference between observed and submitted value.
-
+	ProfitThreshold              int64                 `json:"profitThreshold"`  // Minimum percent of profit when submitting a solution. To avoid loosing money when gas cost is high.
```
themandalore, line 88: add example

pkg/ops/miningManager.go
krasi-georgiev: updated the issue on the tellor core repo as this needs to be added there.
https://github.com/tellor-io/TellorCore/issues/101

pkg/ops/miningManager.go
krasi-georgiev: I updated the logic please double check now. 
With the new logic it just blocks until a transaction is profitable and only then it submits.

One problem I see with this approach is that by the time a transaction is profitable the challenge might have changed and the solution would be invalid, but fixing this would require some greater refactorings and I would say to leave it for another PR. I also added a new issue to simplify the logic in the manager here as I think it can be much more simple than now.
tracking issue https://github.com/tellor-io/TellorMiner/issues/232

pkg/config/config.go
krasi-georgiev: added.

pkg/ops/miningManager.go
```diff
@@ -121,26 +166,188 @@ func (mgr *MiningMgr) Start(ctx context.Context) {
...
+		mgr.log.Debug("saving transaction cost txHash:%v cost GWEI:%v slot:%v", receipt.TxHash.String(), txCost.Int64()/tellorCommon.GWEI, slotNum.Int64())
+		txCostID := tellorCommon.PriceTXs + slotNum.String()
+		_, err = mgr.database.Put(txCostID, txCost.Bytes())
+		if err != nil {
+			mgr.log.Error("saving transaction cost in the db err:%v", err)
+		}
+	}(tx)
+}
+
+func (mgr *MiningMgr) waitProfitable() error {
```
themandalore, line 308: will this exit if a new challenge get's passed down?

krasi-georgiev:
> There seems to be an issue with the submission as some transactions fail with Fail with error 'Request ID is wrong' so please hold it for now.
rinkeby.etherscan.io/tx/0xc73bb449c9a4968a8fd4d4e379981db31dcf58645b3b13612eaf4fa74e23a34d

Further checking on this and after talking to @themandalore this seems normal when another miner has beet you to submitting a chalange.

pkg/ops/miningManager.go
krasi-georgiev: This will block until a SubmitSolution is profitable, 
Checks the current slot, checks the reward and if profitable it will unblock otherwise it waits 1 sec and checks again. Once it unblocks if the solution it submitted is already submitted I assume the transaction will fail and then it will need to calculate the new solution.

themandalore: Yeah, this will need to change. You want it to unblock if a new challenge comes down

krasi-georgiev: Completely refactored the code flow and it is not blocked and I think it is much easier to follow.

Please review only the mining manager changes. The changes regarding the logging are from a different PR and will disappear once that other PR is merged.

pkg/ops/miningManager.go
krasi-georgiev: Completely refactored the code flow and it is not blocked and I think it is much easier to follow.

pkg/ops/miningManager.go
```diff
@@ -77,70 +119,272 @@ func CreateMiningManager(ctx context.Context, exitCh chan os.Signal, submitter t
...
+	for {
+		select {
+		// Boss wants us to quit for the day.
+		case <-mgr.exitCh:
+			mgr.Running = false
+			return
+		// Found a solution.
+		case solution := <-mgr.solutionOutput:
+			// There is no new challenge so resend any pending solution.
+			if solution == nil {
```
themandalore, line 142: when would solution be nil here?

krasi-georgiev: `mgr.tasker.GetWork` sends nil to indicate that the challenge hasn't changed this signals the manager that can use the pendingSolution.

Here:
https://github.com/tellor-io/TellorMiner/pull/228/files?w=1#diff-455c0953444badebf81c8379c2da7d4ad3ba73c891e49ab10a4e13bd2a2a68b5R209

themandalore (approved):

Issue #227, opened 2020-10-26
Investigate why the DB is wiped on every restart. Refactor  to remove the need for this.


krasi-georgiev:
https://github.com/tellor-io/tellorMiner/blob/c13e780e4bf73df8dbe27d61f1ca00d90b3b7f33/cmd/tellor/main.go#L95

This looks strange and shouldn't be needed.

We use the db to save and retrieve transaction costs so deleting the db on every restarts means that every time it will take some time before we know the cost for each transaction.

krasi-georgiev:
@themandalore mentioned that the main reason is that when restarting the miner it will submit the latest available values which mean that these might be quite old and outdated. Another reason is that some trackers submit the 24h average(AMPL) and this needs to be taken into consideration. 

We now keep the transaction costs for each submission slot in the DB so instead of deleting the whole DB I would say a quick fix would be to delete all available values upon restart and leave all nonvalues entries in the DB(e.g. transaction costs)

PR #239, opened 2020-10-28
WIP - Don't delete the db on every restart.

krasi-georgiev:
closes: https://github.com/tellor-io/TellorMiner/issues/227
This looks like some mistake in the early stages of the project. Or because of some subtle bug the couldn't be solved at the time.

My suggestion is to keep the db between restarts is this is crucial for keeping track of the transaction cost and if we notice any problems can fix those in follow up PRs.

### DIFF
--- a/cmd/tellor/main.go
+++ b/cmd/tellor/main.go
@@ -92,7 +92,6 @@ func buildContext() error {
 func AddDBToCtx(remote bool) error {
 	cfg := config.GetConfig()
 	// Create a db instance
-	os.RemoveAll(cfg.DBFile)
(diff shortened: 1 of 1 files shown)


krasi-georgiev:
need to check what happens if the db contains old values upon restart, there is chance to submit old values.

mikeghen:
Can this get into the next release? After it's properly tested of course

krasi-georgiev:
First, need to finish the high priority issues so quite unlikely, I would say most likely after 1-2 releases.

krasi-georgiev:
closing now as not a priority and will revisit at a later stage.

Issue #229, opened 2020-10-26
AMPL Value not Available


mikeghen:
Sometimes the AMPL/USD price isn't available even after 24 hours

themandalore:
help us replicate @mikeghen 

But we know that the 24hr average is still a pain and we're open for Ideas on fixing these

mikeghen:
@themandalore, it's tricky, I first noticed the issue after removing the Kucoin endpoint in mid-October, I added it back after the AMPL price issue last week just to see how it behaved and it worked again. I'm still experimenting in development with it, but I haven't upgraded to the latest version yet, if I find anything useful, I'll post it here

PR #244, opened 2020-11-02
manualData, manualData submission, config changes

themandalore:
fixes: https://github.com/tellor-io/TellorMiner/issues/245


pkg/pow/miningTasker.go
```diff
@@ -132,13 +136,29 @@ func (mt *MiningTasker) GetWork(chan *Work) (*Work, bool) {
 		m2, err := mt.proxy.BatchGet([]string{valKey})
 		if err != nil {
 			mt.log.Info("Could not retrieve pricing data for current request id: %v", err)
-			return nil, false
+			//return nil, false
```
krasi-georgiev, line 139: Why would it continue with an err here? If the db doesn't return even one of the values a solution can't be  submitted right?

pkg/pow/miningTasker.go
```diff
@@ -132,13 +136,29 @@ func (mt *MiningTasker) GetWork(chan *Work) (*Work, bool) {
...
 		}
 		val := m2[valKey]
-		if len(val) == 0 {
+		if val == nil || len(val) == 0 {
```
krasi-georgiev, line 142: When a slice in golang is `nil` then `len(val)` also returns 0 so there is no need to have a double check.

pkg/pow/miningTasker.go
```diff
@@ -132,13 +136,29 @@ func (mt *MiningTasker) GetWork(chan *Work) (*Work, bool) {
...
+			cfg := config.GetConfig()
+			indexPath := filepath.Join(cfg.ConfigFolder, "manualData.json")
```
krasi-georgiev, line 144: But the manual data is already added with the `BuildIndexTrackers` and `configs/indexes.json`

https://github.com/tellor-io/TellorMiner/blob/e97654525c2d34f45979e045311ec9cdf7cf23ad/configs/indexes.json#L147

Since https://github.com/tellor-io/TellorMiner/issues/245 points out that 0 values should not be submitted I think here it should just return an error. The same inside the solution handler.

pkg/pow/miningTasker.go
mikeghen: I agree, I think it should return, I hit this a lot because of #229

pkg/pow/solutionHandler.go
```diff
@@ -54,20 +58,37 @@ func (s *SolutionHandler) Submit(ctx context.Context, result *Result) (*types.Tr
...
-			s.log.Warn("have not retrieved price data for requestId %d. WARNING: Submitting 0 because of faulty API request", challenge.RequestIDs[i].Uint64())
-		}
-		value, err := hexutil.DecodeBig(string(val))
-		if err != nil {
-			if challenge.RequestIDs[i].Uint64() > 53 {
-				s.log.Error("problem decoding price value prior to submitting solution: %v\n", err)
-				if len(val) == 0 {
-					s.log.Error("0 value being submitted")
-					s.currentValues[i] = big.NewInt(0)
+		value := big.NewInt(0)
```
mikeghen: lines like this cost me big before, maybe this should be removed too? Not sure it's needed since you have lines 78 and 80 here, maybe it is though

pkg/pow/miningTasker.go
krasi-georgiev: had a sync call with @themandalore and the problem is that the ovveride works on per symbol basis and not per request ID

pkg/pow/miningTasker.go
```diff
@@ -132,12 +137,32 @@ func (mt *MiningTasker) GetWork(chan *Work) (*Work, bool) {
...
+			if err != nil {
+				fmt.Println("manualData read error", err)
+				return nil, false
+			}
+			defer jsonFile.Close()
+			byteValue, _ := ioutil.ReadAll(jsonFile)
+			var result map[string]map[string]uint
+			err = json.Unmarshal([]byte(byteValue), &result)
+			if err != nil {
+				fmt.Println("manualData unsmarshal ", err)
```
krasi-georgiev, line 156: Lets not use Println anymore :) there is already a logger here.
```suggestion
				mt.log.Error("manualData unsmarshal err:%v", err)
```

krasi-georgiev (approved):
Just few small nits and can merge this workaround.

had a sync call with @themandalore and the main issue is that the override works on per symbol basis and not per request ID

Will track in : https://github.com/tellor-io/TellorMiner/issues/249

PR #248, opened 2020-11-03
Added paid api to config.json

JGcarv:
The idea is to just add a mapping of ` SYMBOL -> apiKey`  in config.json, which will be appended  in the end of the url. Not sure if it works for 100% of apis, but it think it should for most of them.


configs/config.json
```diff
@@ -8,5 +8,8 @@
     "serverPort": 5000,
     "gasMax": 10,
     "dbFile": "/tellorDB",
+    "paidApis": {
+        "vixeod": ["json(https://www.quandl.com/api/v3/datasets/CHRIS/CBOE_VX1.json?api_key=28exxxxxxxxxxxxx).dataset.data.0.4"]
```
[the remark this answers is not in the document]
mikeghen: I concur with @krasi-georgiev, config.json is no place for secrets

configs/indexes.json
```diff
@@ -189,5 +189,8 @@
 	],
 	"ZRX/USD": [
 		"json(https://api.binance.com/api/v1/klines?symbol=ZRXUSDT&interval=1d&limit=1).0.4"
+	],
+	"VIXEOD": [
+
```
mikeghen, line 194: I'm not sure this is a good idea because now you're spreading out where the data is for all indexes, is there any reason the api endpoint can't be here? It seems like there is already custom handling for some indexes, this one seems to only require an API key substitution using environment variable as commented above

mikeghen (commented):
I left some comments, the secret data in the config.json is definitely something that blocks this IMO, my second comment about indexes.json isn't critical, I don't feel strongly about it, just leaving the thought there

configs/indexes.json
themandalore: I think ideally you'd want people to be putting their own in.   We don't want people to think that one specific API is what they're required to pull or use.

krasi-georgiev: Yep, a comment here can explain that the API can be replaced, but I also agree that a single file should be used to specify the APIs, with my comment from that last review I meant to put the API URL here with the following format
`"json(https://www.quandl.com/api/v3/datasets/CHRIS/CBOE_VX1.json?api_key=${VIXEOD_API_KEY}).dataset.data.0.4"` and `VIXEOD_API_KEY` is replaced from the .env file

configs/config.json
```diff
@@ -8,5 +8,8 @@
     "serverPort": 5000,
     "gasMax": 10,
     "dbFile": "/tellorDB",
+    "paidApis": {
+        "VIXEOD": ["json(https://www.quandl.com/api/v3/datasets/CHRIS/CBOE_VX1.json?api_key=${VIXEOD_KEY}).dataset.data.0.4"]
```
themandalore, line 12: note that the name of the key should be "Symbol" + "_KEY" in order for the script to work

configs/indexes.json
themandalore: we don't want everyone using the same API's, so I don't think it's too much of an issue whether you have to manually be updating your config file or your indexes (I personally prefer config since it's shorter)

krasi-georgiev: > we don't want everyone using the same API's,

didn't understand this bit, can you explain. At the moment people use the indexes file and can change it their personal setup and preference. The config for me is to setup how the miner works and the indexes file is to setup how the trackers acquire data. Mixing both breaks the responsibilities separation. I think we should either use the config for everything or keep it separate as it is now - config for the miner setup and indexes for the trackers.


It would be confusing to spread tracker URLs in 2 places. this also add more complexity to the code to handle the paid vs unpaid APIs differently.

krasi-georgiev:
also best to add an entry in the changelog file so that we don't forget it for the next release.

```
## Unreleased

### Added
- [#248](https://github.com/tellor-io/TellorMiner/pull/248) Trackers: Added support for paid APIs and env variable substitutions for the API keys.

```

configs/indexes.json
krasi-georgiev: Here is my idea for this. let me know if you see any problems with this solution.
https://github.com/tellor-io/TellorMiner/pull/254

themandalore:
https://github.com/tellor-io/TellorMiner/pull/254

moving to Krasi's

PR #250, opened 2020-11-04
release set up

themandalore:


JGcarv (approved):

.gitignore
```diff
@@ -8,3 +8,4 @@ pkg/contracts
 
 /tellor
 
+/configs/myconfig*
```
krasi-georgiev, line 11: This looks specific to a local setup so best to keep these excludes in

`.git/info/exclude`
this file it not tracked and can put all excludes specific to someones local env setup.

pkg/pow/solutionHandler.go
```diff
@@ -70,10 +70,7 @@ func (s *SolutionHandler) Submit(ctx context.Context, result *Result) (*types.Tr
 			defer jsonFile.Close()
 			byteValue, _ := ioutil.ReadAll(jsonFile)
 			var result map[string]map[string]uint
-			err = json.Unmarshal([]byte(byteValue), &result)
-			if err != nil {
-				return nil, errors.Wrapf(err, "manualData read Error")
-			}
+			_ = json.Unmarshal([]byte(byteValue), &result)
```
krasi-georgiev, line 73: Why is the error ignored here?

PR #255, opened 2020-11-09
Removes the kucoin api for AMPL as it is not considered stable enough and adds WAVES/USD, OGN/USD, VIXEOD currencies as new requests IDs

themandalore:
Added ID 54,55,56


pkg/tracker/valueGenerators.go
```diff
@@ -96,7 +96,9 @@ func UpdatePSRs(ctx context.Context, updatedSymbols []string) error {
 		bigVal.SetFloat64(amt)
 		bigInt := new(big.Int)
 		bigVal.Int(bigInt)
-
+		// if requestID >= 54 {
+		// 	fmt.Println(requestID, amt)
+		// }
```
krasi-georgiev, line 101:
```suggestion
```

we need to add logging here and enable the debug logs with a flag instead

pkg/pow/solutionHandler.go
```diff
@@ -80,7 +80,7 @@ func (s *SolutionHandler) Submit(ctx context.Context, result *Result) (*types.Tr
 		} else {
 			value, err = hexutil.DecodeBig(string(val))
 			if err != nil {
-				if challenge.RequestIDs[i].Uint64() > 53 {
+				if challenge.RequestIDs[i].Uint64() > 56 {
```
krasi-georgiev, line 83: to avoid changing this all the time we can add in the `tracker` package
```
func MaxPSRID() uint64 {
	var maxID int
	for id := range PSRs {
		if id > maxID {
			maxID = id
		}
	}
	return uint64(maxID)
}
```
and then this will become

```suggestion
				if challenge.RequestIDs[i].Uint64() > tracker.MaxPSRID() {
```

btw can you double check the logic here as it seems strange, why does it run this check only when there is an error?

themandalore: If you have a value, the requestID obviously isn't above the max psr id...so not need checking

krasi-georgiev: Yep, I see your point, but we still need to make the errors and checks a bit more clear, can be a diff PR, but should have it in mind.

krasi-georgiev (approved):

PR #262, opened 2020-11-11
Add Solidity syntax highlight

cleanunicorn:
Update `.gitattributes` to enable Solidity syntax highlight on GitHub.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.sol linguist-language=Solidity


krasi-georgiev:
Thanks. I just a quick look and it seems that github detects this by default

https://github.com/github/linguist

Why do you think it needs to be added?

krasi-georgiev:
I don't think we need this as github already detects the .sol files and does the proper highlighting, but I will let @cleanunicorn give some more context.

krasi-georgiev:
@cleanunicorn closing this now, but feel free to reopen with some more context on why you think we should add this.
Thanks

PR #274, opened 2020-11-13
partially removing ctx from balanceCmd

JGcarv:


cmd/tellor/main.go
```diff
@@ -156,7 +156,8 @@ func App() *cli.Cli {
 	app.Command("stake", "staking operations", stakeCmd(logSetup, logLevel))
 	app.Command("transfer", "send TRB to address", moveCmd(ops.Transfer, logSetup, logLevel))
 	app.Command("approve", "approve TRB to address", moveCmd(ops.Approve, logSetup, logLevel))
-	app.Command("balance", "check balance of address", balanceCmd)
+	//Using values from context, until we have a function that setups the client and returns as values, not as part of the context
```
krasi-georgiev, line 159:
```suggestion
	// Using values from context, until we have a function that setups the client and returns as values, not as part of the context.
```
Golang commentary guidelines.

krasi-georgiev (commented):
Yep that is a very good start we can even merge it like this and you can continue in the same direction.

pkg/ops/transferOps.go
```diff
@@ -75,16 +75,13 @@ func Approve(ctx context.Context, logger log.Logger, _spender common.Address, am
 	return nil
 }
 
-func Balance(ctx context.Context, addr common.Address) error {
-	client := ctx.Value(tellorCommon.ClientContextKey).(rpc.ETHClient)
-
+func Balance(ctx context.Context, client rpc.ETHClient, getterInstance *getter.TellorGetters, addr common.Address) error {
+	//client.BalanceAt doesn't need the context, will remove later since it's used elsewhere
```
krasi-georgiev, line 79: The `ctx context.Context` in  `client.BalanceAt(context.Background()` is used properly. It is for cancelling a request, This is the main purpose of a context - you can cancel it manually - for example if you want to exit the application(by calling context cancel) without waiting for a reply or can set it to auto-cancel for example if you want to set a timeout of lets say 1sec.

krasi-georgiev: in our case we will use the context to be able to stop pending eth client requests when a SIGTERM is sent to the app.

pkg/ops/transferOps.go
```diff
@@ -75,16 +75,12 @@ func Approve(ctx context.Context, logger log.Logger, _spender common.Address, am
 	return nil
 }
 
-func Balance(ctx context.Context, addr common.Address) error {
-	client := ctx.Value(tellorCommon.ClientContextKey).(rpc.ETHClient)
-
+func Balance(ctx context.Context, client rpc.ETHClient, getterInstance *getter.TellorGetters, addr common.Address) error {
 	ethBalance, err := client.BalanceAt(context.Background(), addr, nil)
```
krasi-georgiev, line 79:
```suggestion
	ethBalance, err := client.BalanceAt(ctx, addr, nil)
```

The idea is that here the context will be created in the main package and then passed to the this component, this way the main package can cancel the context and this will propagate here and will cancel any pending requests and the function will return. Withoug this there is no way for the main package to control when the child function/components need to exit.

Here `context.Background()` creates a new empty context, but instead we should just use the one provided in the `Balance(ctx context.Context, `

krasi-georgiev (commented):
Thanks. Since we still don't have e2e tests for this, can you please also test the `balanceCmd` locally.

cmd/tellor/main.go
```diff
@@ -193,16 +193,22 @@ func moveCmd(f func(context.Context, log.Logger, common.Address, *big.Int) error
...
 func balanceCmd(cmd *cli.Cmd) {
+
 	addr := ETHAddress{}
 	cmd.VarArg("ADDRESS", &addr, "ethereum public address")
 	cmd.Spec = "[ADDRESS]"
 	cmd.Action = func() {
+		// Using values from context, until we have a function that setups the client and returns as values, not as part of the context
+		getter := ctx.Value(tellorCommon.ContractsGetterContextKey).(*getter.TellorGetters)
+		client := ctx.Value(tellorCommon.ClientContextKey).(rpc.ETHClient)
+		commonAddress := ctx.Value(tellorCommon.PublicAddress).(common.Address)
```
JGcarv, line 204: Due to the order the functions were called, the ctx is emprty until cmd.Action() is callled, so this is the place where it needs to be to work as is. I think that's not very good though, as we need a similar structure for all commands. It need to be refactored to account for that. Maybe include that when changing de cli paackage?

krasi-georgiev: I am ok with that for the sake to continue the refactoring.

If you confirm that this works when testing locally I am ok to merge it.

JGcarv: It's working locally!

krasi-georgiev (approved):

PR #277, opened 2020-11-16
Make the docker repo name dinamicaly linked to the gh repo name

krasi-georgiev:
As mentioned in the chat lets move towards renaming it to just tellor as this cli does more then just mining - this is the tellor cli tool - 

you can run it as a miner
only as a values submitter
for voting
for opening disputes
....


krasi-georgiev:
hold it for now untill we decide on the name for the cli

krasi-georgiev:
I added the image name to be linked to the github repo  so the docker repo will always match the github repo name

Issue #236, opened 2020-10-28
Replace all `fmt.Errorf` with `errors.Wrap`  and `errors.Errorf` 


krasi-georgiev:

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 10.0779 TRB (281.17 USD @ $27.9/TRB)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/tellor-io/TellorMiner/236/100024080).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/tellor-io/TellorMiner/236/100024080).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://chat.gitcoin.co/'>Gitcoin Chat</a>
* $1,052,111.78 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

developerfred:
@themandalore 
I'm interested in working on this issue, invite me through Gitcoin

my work plan is to make all the changes

themandalore:
@developerfred , sent.  You're our space monkey with regards to testing out this whole gitcoin thing, so let me know if you need anything

developerfred:
@themandalore Thanks, perfect!

themandalore:
Looks like there's still some fmt.Errorf in httpRetriever.go and logConfig.go (the util package)

developerfred:
@themandalore I'll check here, thanks for letting me know.

developerfred:
@themandalore done.

themandalore:
Looks good, let me know the next steps with regard to gitcoin

developerfred:
@themandalore You can add me to the bounty, when I try to express interest I get an error. I reported to the team and I'm analyzing the Gitcoin code to see why. 

my gitcoin user: @developerfred

themandalore:
@developerfred , I can't really do anything. I'll reach out to them, but if it takes more than a day or two, just reach back out and we'll close it and manually fix it

developerfred:
@themandalore perfect, i'm talking to them too. 
They will fix it now, then the process is just to approve me and then pay.

developerfred:
@themandalore done, now you need to approve me on Gitcoin, for me to submit PR.

krasi-georgiev:
[...]

PR #286, opened 2020-11-19
 change errors.Errorf  to *.Wrapf and *.Wrap

developerfred:
fixes: #236

Fix
https://github.com/tellor-io/TellorMiner/issues/236#issuecomment-729972707

### DIFF
--- a/cmd/tellor/argTypes.go
+++ b/cmd/tellor/argTypes.go
@@ -41,7 +41,7 @@ type ETHAddress struct {
 func (a *ETHAddress) Set(v string) error {
 	valid := common.IsHexAddress(v)
 	if !valid {
-		return errors.Errorf("%s is not a valid etherum address format", v)
+		return errors.Errorf("%s valid etherum address format", v)
 	}
 	a.addr = common.HexToAddress(v)
 	return nil
--- a/cmd/tellor/argTypes.go
+++ b/cmd/tellor/argTypes.go
@@ -41,7 +41,7 @@ type ETHAddress struct {
 func (a *ETHAddress) Set(v string) error {
 	valid := common.IsHexAddress(v)
 	if !valid {
-		return errors.Errorf("%s valid etherum address format", v)
+		return errors.Errorf("invalid etherum address:%v", v)
 	}
 	a.addr = common.HexToAddress(v)
 	return nil
@@ -63,10 +63,10 @@ func (b *EthereumInt) Set(v string) error {
 	g := new(big.Int)
 	_, ok := g.SetString(v, 10)
 	if !ok {
-		return errors.Errorf("%s is not a valid integer", v)
+		return errors.Errorf("invalid integer:%v", v)
 	}
 	if len(g.Bytes()) > 32 {
-		return errors.Errorf("%s is larger than 256 bits", v)
+		return errors.Errorf("invalid size larger than 256 bits:%v", v)
 	}
 	b.Int = g
 	return nil
--- a/pkg/apiOracle/valueOracle.go
+++ b/pkg/apiOracle/valueOracle.go
@@ -110,14 +110,14 @@ func EnsureValueOracle() error {
 		if os.IsNotExist(err) {
 			exists = false
 		} else {
-			return errors.Errorf("file %s stat error: %v", historyPath, err)
+			return errors.Wrapf(err, "stats for file: %v", historyPath)
 		}
 	}
 
 	if exists {
 		byteValue, err := ioutil.ReadFile(historyPath)
 		if err != nil {
-			return errors.Errorf("failed to read psr file @ %s: %v", historyPath, err)
+			return errors.Wrapf(err, "failed to read psr file @: %v", historyPath)
 		}
 		err = json.Unmarshal(byteValue, &valueHistory)
 		if err != nil {
--- a/pkg/apiOracle/valueOracle.go
+++ b/pkg/apiOracle/valueOracle.go
@@ -13,7 +13,10 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/tellor-io/TellorMiner/pkg/config"
+	"github.com/tellor-io/TellorMiner/pkg/util"
 )
+
+var logger = util.NewLogger("apiOracle", "valueOracle")
 
 // maps symbol to a time window of values.
 var valueHistory map[string]*Window
@@ -64,7 +67,8 @@ func writeOutHistory() {
 	// this function is single threaded, but we need mutex to access multithreaded history.
 	valueHistoryMutex.Unlock()
 	if err != nil {
-		errors.Wrap(err, "marshal PSR values")
+		logger.Error("failed to marshal PSR values: %s", err.Error())
+		return
 	}
 
 	cfg := config.GetConfig()
@@ -72,12 +76,14 @@ func writeOutHistory() {
 	psrSavedDataTmp := psrSavedData + ".tmp"
 	err = ioutil.WriteFile(psrSavedDataTmp, data, 0644)
 	if err != nil {
-		errors.Wrapf(err, "write out PSR values to: %s", psrSavedDataTmp)
+		logger.Error("failed to write out PSR values to %s: %s", psrSavedDataTmp, err.Error())
+		return
 	}
 	// Rename tmp file to old file (should be atomic on most modern OS)
 	err = os.Rename(psrSavedDataTmp, psrSavedData)
 	if err != nil {
-		errors.Wrap(err, "move new PSR save onto old")
+		logger.Error("failed move new PSR save onto old: %s", err.Error())
+		return
 	}
 }
 
@@ -104,18 +110,18 @@ func EnsureValueOracle() error {
 		if os.IsNotExist(err) {
 			exists = false
 		} else {
-			return errors.Wrapf(err, "stats for file:%v", historyPath)
+			return errors.Wrapf(err, "stat error file: %v", historyPath)
 		}
 	}
 
 	if exists {
 		byteValue, err := ioutil.ReadFile(historyPath)
 		if err != nil {
-			return errors.Wrapf(err, "read psr file @:%v", historyPath)
+			return errors.Wrapf(err, "read psr file:%v", historyPath)
 		}
 		err = json.Unmarshal(byteValue, &valueHistory)
 		if err != nil {
-			return errors.Wrap(err, "failed to unmarshal saved")
+			return errors.Errorf("failed to unmarshal saved")
 		}
 	} else {
 		valueHistory = make(map[string]*Window)
--- a/pkg/apiOracle/valueOracle.go
+++ b/pkg/apiOracle/valueOracle.go
@@ -13,10 +13,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/tellor-io/TellorMiner/pkg/config"
-	"github.com/tellor-io/TellorMiner/pkg/util"
 )
-
-var logger = util.NewLogger("apiOracle", "valueOracle")
 
 // maps symbol to a time window of values.
 var valueHistory map[string]*Window
@@ -68,7 +65,6 @@ func writeOutHistory() {
 	valueHistoryMutex.Unlock()
 	if err != nil {
 		errors.Wrap(err, "marshal PSR values")
-		return
 	}
 
 	cfg := config.GetConfig()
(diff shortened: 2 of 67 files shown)


cmd/tellor/argTypes.go
```diff
@@ -41,7 +41,7 @@ type ETHAddress struct {
 func (a *ETHAddress) Set(v string) error {
 	valid := common.IsHexAddress(v)
 	if !valid {
-		return errors.Errorf("%s is not a valid etherum address format", v)
+		return errors.Errorf("%s valid etherum address format", v)
```
krasi-georgiev, line 44:
```suggestion
		return errors.Errorf("invalid etherum address:%v", v)
```

It is more structured when we have the param at the end.

pkg/apiOracle/valueOracle.go
```diff
@@ -110,14 +110,14 @@ func EnsureValueOracle() error {
...
-			return errors.Errorf("file %s stat error: %v", historyPath, err)
+			return errors.Wrapf(err, "stats for file: %v", historyPath)
 		}
 	}
 
 	if exists {
 		byteValue, err := ioutil.ReadFile(historyPath)
 		if err != nil {
-			return errors.Errorf("failed to read psr file @ %s: %v", historyPath, err)
+			return errors.Wrapf(err, "failed to read psr file @: %v", historyPath)
```
krasi-georgiev, line 120:
```suggestion
			return errors.Wrapf(err, "read psr file @: %v", historyPath)
```

pkg/config/config.go
```diff
@@ -132,7 +132,7 @@ const PrivateKeyEnvName = "ETH_PRIVATE_KEY"
 func ParseConfig(path string) error {
 	data, err := ioutil.ReadFile(path)
 	if err != nil {
-		return errors.Errorf("failed to open config file %s: %v", path, err)
+		return errors.Wrapf(err, "to open config file: %v", path)
```
krasi-georgiev, line 135:
```suggestion
		return errors.Wrapf(err, "open config file: %v", path)
```

pkg/config/config.go
```diff
@@ -141,7 +141,7 @@ func ParseConfig(path string) error {
 func ParseConfigBytes(data []byte) error {
 	err := json.Unmarshal(data, &config)
 	if err != nil {
-		return errors.Errorf("failed to parse json: %s", err.Error())
+		return errors.Wrap(err, "to parse json")
```
krasi-georgiev, line 144:
```suggestion
		return errors.Wrap(err, "parse json")
```

pkg/config/config.go
```diff
@@ -169,15 +169,15 @@ func ParseConfigBytes(data []byte) error {
 
 	err = validateConfig(&config)
 	if err != nil {
-		return errors.Errorf("validation failed: %s", err)
+		return errors.Wrap(err, "validation")
```
krasi-georgiev, line 172:
```suggestion
		return errors.Wrap(err, "config validation")
```

pkg/config/config.go
```diff
@@ -189,18 +189,18 @@ func validateConfig(cfg *Config) error {
 	} else {
 		b, err = hex.DecodeString(cfg.PrivateKey)
 		if err != nil || len(b) != 32 {
-			return errors.Errorf("expecting 64 hex character private key, got \"%s\"", cfg.PrivateKey)
+			return errors.Wrapf(err, "expecting 64 hex character private key, got \"%s\"", cfg.PrivateKey)
 		}
 		if len(cfg.ContractAddress) != 42 {
-			return errors.Errorf("expecting 40 hex character contract address, got \"%s\"", cfg.ContractAddress)
+			return errors.Wrapf(err, "expecting 40 hex character contract address, got \"%s\"", cfg.ContractAddress)
```
krasi-georgiev, line 195:
```suggestion
			return errors.Errorf("expecting 40 hex character contract address, got \"%s\"", cfg.ContractAddress)
```
There is no error to wrap here. You are printing the error from the last if block

The same in few other places

krasi-georgiev (commented):
Thanks for the update, left few comments, please have a look at all other occurrences and continue in the same direction.

developerfred:
@krasi-georgiev  thanks for review! I make changes.

krasi-georgiev:
please change the rest of the PR according to my comments as these changes apply in many different places not just the ones I pointed.

krasi-georgiev:
ping when ready for another review or if you get stuck making all tests pass.

cmd/tellor/argTypes.go
```diff
@@ -63,10 +63,10 @@ func (b *EthereumInt) Set(v string) error {
 	g := new(big.Int)
 	_, ok := g.SetString(v, 10)
 	if !ok {
-		return errors.Errorf("%s is not a valid integer", v)
+		return errors.Errorf("invalid integer: %v", v)
 	}
 	if len(g.Bytes()) > 32 {
-		return errors.Errorf("%s is larger than 256 bits", v)
+		return errors.Errorf("invalid size: %v is larger than 256 bits", v)
```
krasi-georgiev, line 69:
```suggestion
		return errors.Errorf("invalid size larger than 256 bits:%v", v)
```
it is common practise to put arguments at the end.

pkg/db/remoteRequest.go
```diff
@@ -72,7 +72,7 @@ func createRequest(dbKeys []string, values [][]byte, signer RequestSigner) (*req
 	}
 	if sig == nil {
 		log.Error("Signature was not generated")
```
krasi-georgiev, line 74:
```suggestion
```
not related to the PR , but we can remove this log entry, no point on having an error and a log, it is kind of duplicate.

pkg/db/remoteRequest.go
```diff
@@ -72,7 +72,7 @@ func createRequest(dbKeys []string, values [][]byte, signer RequestSigner) (*req
...
-		return nil, errors.Errorf("Could not generate a signature for  hash: %v", hash)
+		return nil, errors.Wrapf(err, "Could not generate a signature for  hash: %v", hash)
```
krasi-georgiev, line 75:
```suggestion
		return nil, errors.Wrapf(err, "generate a signature for  hash: %v", hash)
```
we don't need words like this, just need to describe when the error occurred. It is obvious that there is `couldn't` :smile: 

Also, all errors are lower case and no full stop at the end.

developerfred:
@krasi-georgiev  double check in files and test pass done.

pkg/ops/stakeOps.go
```diff
@@ -74,24 +74,24 @@ func Deposit(ctx context.Context, logger log.Logger) error {
 	copy(dat32[:], dat)
 	stakeAmt, err := tmaster.GetUintVar(nil, dat32)
 	if err != nil {
-		return errors.Errorf("fetching stake amount failed: %s", err.Error())
+		return errors.Wrap(err, "fetching stake amount")
 	}
 
 	if balance.Cmp(stakeAmt) < 0 {
-		return errors.Errorf("insufficient balance (%s), mining stake requires %s TRB",
+		return errors.Wrapf(err, "insufficient balance (%s), mining stake requires %s TRB",
```
krasi-georgiev, line 81: Where is this error from? Please be careful about such cases as with this the error will be masked and not even returned.

krasi-georgiev (commented):
We are going in a good direction, thanks.
Please review my comments and check the rest of the pr to make sure you apply the same principles everywhere.

pkg/ops/stakeOps.go
developerfred: > Where is this error from? Please be careful about such cases as with this the error will be masked and not even returned.

I will be careful. thanks.

scripts/opencl/opencl.go
```diff
@@ -36,28 +36,28 @@ func compileSources() (string, error) {
...
-		return "", errors.Errorf("missing %s file", headerFile)
+		return "", errors.Wrapf(err, "missing %s file", headerFile)
 	}
 	out.Write(contents[headerFile])
 	delete(contents, headerFile)
 
 	mainSrc, ok := contents[kernelFile]
 	if !ok {
-		return "", errors.Errorf("missing %s file", kernelFile)
+		return "", errors.Wrapf(err, "missing %s file", kernelFile)
```
krasi-georgiev, line 60: Where is this error coming from? Please double-check the entire PR for such issues.

developerfred: When I didn't have a declared error variable I put `errors.Errorf(msg)`
I will change it to `Wrapf` it is better to have control of errors.

developerfred: I put missing kernel and missing header

krasi-georgiev: The question still stands, where does this `err` variable come from?

scripts/opencl/opencl.go
```diff
@@ -36,28 +36,28 @@ func compileSources() (string, error) {
...
 		if err != nil {
-			return "", errors.Errorf("couldn't read file %s: %s", filepath, err.Error())
+			return "", errors.Wrapf(err, "read file: %s", filepath)
 		}
 		contents[file.Name()] = data
 	}
 	var out strings.Builder
 	if _, ok := contents[headerFile]; !ok {
-		return "", errors.Errorf("missing %s file", headerFile)
+		return "", errors.Wrapf(err, "missing header: %s file", headerFile)
```
krasi-georgiev, line 53: The question still stands. Where does this `err` variable come from?

developerfred: ah got it, when there is no variable defined `err` in the function I can use `errors.ErrorF`

another thing i changed some err.Error () to errors lib is it prudent to do this?

krasi-georgiev:
for the error you can run `make lint` locally to find out why it happens

krasi-georgiev:
btw in few minutes will merge another big PR that does some other global refactorings so there will be some conflict to be resolved and unfortunately we can't avoid it, but shouldn't be too difficult to resolve.

scripts/opencl/opencl.go
```diff
@@ -36,28 +35,28 @@ func compileSources() (string, error) {
...
-		return "", errors.Errorf("missing %s file", headerFile)
+		return "", errors.Errorf("missing header: %s file", headerFile)
 	}
 	out.Write(contents[headerFile])
 	delete(contents, headerFile)
 
 	mainSrc, ok := contents[kernelFile]
 	if !ok {
-		return "", errors.Errorf("missing %s file", kernelFile)
+		return "", errors.Errorf("missing kearnal: %s file", kernelFile)
```
krasi-georgiev, line 59:
```suggestion
		return "", errors.Errorf("missing kernel file:%s", kernelFile)
```
please put arguments at the end to make these easier to read, the same in few other places.

pkg/util/logConfig.go
```diff
@@ -32,10 +32,10 @@ func ParseLoggingConfig(file string) error {
 	if len(file) > 0 {
 		info, err := os.Stat(file)
 		if os.IsNotExist(err) {
-			return errors.Errorf("loggingConfigPath references an invalid file at: %s", file)
+			return errors.Wrapf(err, "loggingConfigPath references an invalid file at: %s", file)
 		}
 		if info.IsDir() {
-			return errors.Errorf("logging config file %s is a directory", file)
+			return errors.Wrapf(err, "logging config file %s is a directory", file)
```
krasi-georgiev, line 38: Again this error is set somewhere else so should be used here. Please check the entire PR for these again.

pkg/tracker/currentVariables.go
```diff
@@ -90,7 +90,7 @@ func (b *CurrentVariablesTracker) Exec(ctx context.Context) error {
 
 	err = DB.Put(db.TotalTipKey, []byte(hexutil.EncodeBig(returnNewVariables.Tip)))
 	if err != nil {
-		return errors.Wrap(err, "totaltip put")
+		return errors.Wrap(err, "taltip put")
```
krasi-georgiev, line 93:
```suggestion
		return errors.Wrap(err, "total tip put")
```

pkg/pow/solutionHandler.go
```diff
@@ -89,7 +89,7 @@ func (s *SolutionHandler) Submit(ctx context.Context, result *Result) (*types.Tr
 					}
 					continue
 				}
-				return nil, errors.Errorf("no value in database,  reg id: %v", challenge.RequestIDs[i].Uint64())
+				return nil, errors.Wrapf(err, "no value in database,  reg id:%v", challenge.RequestIDs[i].Uint64())
```
krasi-georgiev, line 92: Where is this error from?
```suggestion
				return nil, errors.Errorf("no value in database,  reg id:%v", challenge.RequestIDs[i].Uint64())
```

krasi-georgiev (approved):

krasi-georgiev:
Thanks!

PR #288, opened 2020-11-19
Removing the usafe of global context

JGcarv:
To avoid functions with 7+ parameters, I created structs for the main pieces previously used in the contex.

Otherwise, the last big piece that we still rely on the ctx is the DB, which I'll tackle next.

### DIFF
--- a/cmd/tellor/main.go
+++ b/cmd/tellor/main.go
@@ -52,9 +52,6 @@ func setup() error {
 			return errors.Wrap(err, "create client instance")
 		}
 
-		// Leaving the client in ctx for now because it's still used in some places
-		ctx = context.WithValue(context.Background(), tellorCommon.ClientContextKey, client)
-
 		// Create an instance of the tellor master contract for on-chain interactions
 		contractAddress := common.HexToAddress(cfg.ContractAddress)
 		contractTellorInstance, err := tellor.NewTellor(contractAddress, client)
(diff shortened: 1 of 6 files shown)


cmd/tellor/main.go
```diff
@@ -60,7 +64,7 @@ func buildContext() error {
 		if err != nil {
 			return errors.Wrap(err, "create tellor transactor instance")
 		}
-
+		// Leaving those in because are still used in some places(miner submission mostly)
```
krasi-georgiev, line 67:
```suggestion
		// Leaving those in because are still used in some places(miner submission mostly).
```

cmd/tellor/main.go
```diff
@@ -31,6 +31,9 @@ import (
 )
 
 var ctx context.Context
+var cont tellorCommon.Contract
+var acc tellorCommon.Account
+var clt rpc.ETHClient
```
krasi-georgiev, line 36: Having global variables is another anti pattern and we need to move away from it. It is in the coding guidelines :smiley:  https://github.com/tellor-io/TellorMiner/blob/master/docs/coding-style-guide.md

There are many articles why global variables are bad.

cmd/tellor/main.go
```diff
@@ -172,41 +181,38 @@ func stakeCmd(logSetup func(string) log.Logger, logLevel *string) func(*cli.Cmd)
 	}
 }
 
-func simpleCmd(f func(context.Context, log.Logger) error, logSetup func(string) log.Logger, logLevel *string) func(*cli.Cmd) {
+func simpleCmd(f func(context.Context, log.Logger, rpc.ETHClient, tellorCommon.Contract, tellorCommon.Account) error, logSetup func(string) log.Logger, logLevel *string) func(*cli.Cmd) {
 	return func(cmd *cli.Cmd) {
 		cmd.Action = func() {
-			ExitOnError(f(ctx, logSetup(*logLevel)), "")
+			ExitOnError(f(ctx, logSetup(*logLevel), clt, cont, acc), "")
```
krasi-georgiev, line 187: These need to be passed as arguments instead of using some global variables.

cmd/tellor/main.go
```diff
@@ -172,41 +181,38 @@ func stakeCmd(logSetup func(string) log.Logger, logLevel *string) func(*cli.Cmd)
...
 		}
 	}
 }
 
-func moveCmd(f func(context.Context, log.Logger, common.Address, *big.Int) error, logSetup func(string) log.Logger, logLevel *string) func(*cli.Cmd) {
+func moveCmd(f func(context.Context, log.Logger, rpc.ETHClient, tellorCommon.Contract, tellorCommon.Account, common.Address, *big.Int) error, logSetup func(string) log.Logger, logLevel *string) func(*cli.Cmd) {
```
krasi-georgiev, line 192:
```suggestion
func moveCmd(
	f func(
		context.Context,
		log.Logger,
		rpc.ETHClient,
		tellorCommon.Contract,
		tellorCommon.Account,
		common.Address, *big.Int,
	) error,
	logSetup func(string) log.Logger,
	logLevel *string,
) func(*cli.Cmd)
```

We will probably refactor this later, but for now can add new lines to make it a little bit more readable

krasi-georgiev (commented):
Yep we are going in the right direction, just refactor to removethe use of global variables and add new lines to the funcs with many arguments to make it a bit more readable and that is it for this one.

krasi-georgiev (approved):

Issue #268, opened 2020-11-12
Convert the docs into gitbook format and create a GH actions daily task to copy all docs from the miner repo into the main tellor docs repo.


krasi-georgiev:
At the moment the miner repo has some docs under `docs` folder and others are kept in the main docs repo at https://github.com/tellor-io/TellorDocs. I think the main docs repo should keep docs related to the project itself and an overview of all components and  miner specific docs(how to build, how to run etc) should be kept here in the miner's repo. The main docs repo will then just have a link to the miners github book docs.
This has 2 main benefits
 - when we make changes to the code in the same PR can also update the docs.
 - easier to keep in sync since there will be no need to open 2 PRs.
 - some parts of the docs can be auto-generated from the code itself: https://github.com/tellor-io/TellorMiner/issues/194

Follow the issue
A solid command of the English language is important for this issue

krasi-georgiev:
On a second thought a much nicer user experience would be to keep all docs in a single github book rather then making people jump between books so to achieve this we should:
 - still keep all miner related docs here in the miner repo(easier to keep the docs in sync with the code changes)
 - create a daily GH actions cron task to copy the miner docs from here to the main tellor docs repo.

I am renaming the title to this new idea.

krasi-georgiev:
this GH action seems to be the one to use
https://github.com/repo-sync/repo-sync

krasi-georgiev:
so the idea is:
 1. one PR to move all docs from `https://github.com/tellor-io/TellorDocs/tree/master/miner-documentation` to this repo in the `/docs/miner-documentation` folder. 
 1. Another PR to merge the docs from `/docs/miner-documentation` with the existing docs in `/docs` and convert them into the git book format.
 1. Another PR to setup `repo-sync/repo-sync` to sync the `docs` folder from this repo into `https://github.com/tellor-io/TellorDocs/tree/master/miner-documentation`
 1. Another PR to add notes some quite visible warnings to never edit `https://github.com/tellor-io/TellorDocs/tree/master/miner-documentation` directly and edit the docs in this repo instead which will be auto-synced with the main docs repo.

With this workflow, it will be quite easy to keep the docs up to date, but still provide a single website to the user to read the entire document.

This needs to be done after or together with https://github.com/tellor-io/TellorMiner/issues/282

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

[...]

PR #298, opened 2020-11-27
merge docs from TellorDocs

hhio618:
### Description
This PR moves all docs from `https://github.com/tellor-io/TellorDocs/tree/master/miner-documentation` to this repo in the `/docs` folder and convert them to the Gitbook format.
Also, I must note INDEX.md file. We'll use INDEX.md for the repo sync process in another PR.
### Refs 
#268

### DIFF
--- a/docs/INDEX.md
+++ b/docs/INDEX.md
@@ -1,0 +1,8 @@
+* [Become a Miner](miner.md)
+* [The Guide](the-guide.md)
+* [Configuation](configuation.md)
+* [Contributing](contributing.md)
+* [Coding style guide](coding-style-guide.md)
+* [Release process](release-process.md)
+* [Reference](miner-reference.md)
+* [Disclaimer](disclaimer.md)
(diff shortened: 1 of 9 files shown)


krasi-georgiev:
could you run `make lint` locally to remove the extra white space and add another commit?

Are these an exact copy from the main doc repo?
Do you if there is any way to preview how would these look like in gitbook?
There is a bit of work on fixing some of the text, but will do this in another PR, lets just have the initial structure right and will worry about the actual test in a different PR.

hhio618:
You're welcome!
> Are these an exact copy from the main doc repo?


I copied these docs from the main doc repo, but the `make lint` command makes some changes to these files, so there is a bit of difference now (removing white noises).





> There is a bit of work on fixing some of the text, but will do this in another PR, lets just have the initial structure and setup the sync and will worry about the actual text in a different PR.
> We have renamed the `miner` into `telliot` - the tellor.io tasker as it can do more than just mining so need to refactor the docs to reflect that but will do it in a follow up PR

Ok, no problem

krasi-georgiev (approved):

PR #305, opened 2020-12-02
Arrange the docs by filename and merge few files together

hhio618:
### Description
Arrange documents in the `docs` folder.
### Refs
#268


docs/04_contributing.md
```diff
@@ -59,7 +59,7 @@ $ <Iterate your development>
 $ git push fork <your_PR_branch>
 ```
 
-1. If your change affects users (adds,removes or changes a feature) add the item to the [CHANGELOG](CHANGELOG.md).
+1. If your change affects users (adds,removes or changes a feature) add the item to the [CHANGELOG](08_CHANGELOG.md).
```
krasi-georgiev, line 62: Hm, I think we should keep the CHANGELOG filename unchanged. Ideally we want it last in the docs list, but it is ok even if shows at the begining.

krasi-georgiev:
Since we are arranging the docs already, can you do these changes as well:
 - merge pages: reference + configuration
 - put the text from Disclaimer into Become a miner
 - in README - change the formating of the block DISCLAIMER so that it doesn't have a vertical scrolling.
 - rename The Guide -> Setup and usage
 - Merge Become a Minero into Setup and usage.

PR #316, opened 2020-12-08
WIP - Onchaincalls

themandalore:
closes https://github.com/tellor-io/telliot/issues/280

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -79,6 +79,7 @@ type Config struct {
 	Worker                       string                `json:"worker"`
 	Password                     string                `json:"password"`
 	PoolURL                      string                `json:"poolURL"`
+	MainnetRPC					 string				   `json: mainnetRPC`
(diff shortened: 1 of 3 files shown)


pkg/config/config.go
```diff
@@ -151,7 +152,9 @@ func ParseConfigBytes(data []byte) error {
 			return errors.Errorf("missing ethereum wallet private key environment variable '%v'", PrivateKeyEnvName)
 		}
 	}
-
+	if len(config.MainnetRPC) ==0{
+		config.MainnetRPC = config.NodeURL;
```
krasi-georgiev, line 156: maybe to avoid maintaining 2 different configs can just add

config.NodeURLMainnet;
config.NodeURLTestnet;

krasi-georgiev: Now I see that many settings are different for mainnet vs testnet so this will probably bloat the settings file.

PR #318, opened 2020-12-11
Migrate indexes.json to JSONPath format

### LINKED ISSUE #221: Use an existing jsopath golang package to remove the need of a custom parser
There is some strange parser and a file format for the indexes.
Lets review it again and try to simplify. Should also remove the need for this special parser and just use json.Marshal/Unmarshal

`
"BNB/USD": [
		"json(https://dex.binance.org/api/v1/klines?symbol=BNB_USDSB-1AC&interval=1d&limit=1).0.4",
`

hhio618:
### Description
Implemented using: https://github.com/yalp/jsonpath
### Refs
Fixes #221


pkg/tracker/index.go
```diff
@@ -165,13 +170,27 @@ func (i *IndexTracker) Exec(ctx context.Context) error {
...
+	if !ok {
+		return errors.Wrap(err, "JsonPath match need to be an array")
+	}
+	for _, a := range arr {
+		val, _ := strconv.ParseFloat(fmt.Sprintf("%v", a), 64)
+		vals = append(vals, val)
+	}
 
-	//fmt.Printf("got value of %f for %s\n", val, i.Identifier )
+	fmt.Fprintln(os.Stdout, fmt.Sprintf("got value of %f for %s\n", vals, i.Identifier))
```
krasi-georgiev, line 193: please use debug logging if you think that this is useful for debugging, otherwise please remove.

krasi-georgiev (commented):
I don't see any removal of the custom parser from 
`github.com/tellor-io/telliot/pkg/util`

This change is relevant to the end-user to need an entry in the changelog.

I think we can even remove `"type": "json"` and instead check that if `jsonPath` is not empty it can assume that the results is `"type": "json"`

We also need a test for this parsing.

pkg/tracker/index.go
hhio618: Used for testing, I removed this now

pkg/tracker/index_test.go
```diff
@@ -3,40 +3,38 @@
...
-// 	if val == nil {
-// 		testutil.Ok(t, errors.New(Expected a value stored for request ID 1))
-// 	}
-// 	intVal, err := hexutil.DecodeBig(string(val))
-// 	if err != nil {
-// 		testutil.Ok(t, err)
-// 	}
-// 	fmt.Println("DB value", intVal)
-// }
+func TestPSR(t *testing.T) {
```
krasi-georgiev, line 14: Actually, we disabled this one because it was flaky. Some APIs have rate limits.

Instead of this, I imagine the following:

Move the config parsing into a separate function,
Parse the config and build an array `[]IndexTracker` for `source` in `IndexTracker` can have some mock `DataSource` that returns something we can test.

Wi

And then can call each tracker in the array `i.DataSource.Get` with the result can call `i.parsePayload(results)` and check expected vs actual result.

pkg/tracker/index.go
```diff
@@ -165,13 +170,32 @@ func (i *IndexTracker) Exec(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	vals := make([]float64, 0)
+
+	// Query the json payload using JSONPath expression
```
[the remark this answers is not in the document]
hhio618: Nice tips, thanks

pkg/tracker/index.go
```diff
@@ -60,50 +61,49 @@ func BuildIndexTrackers() ([]Tracker, error) {
...
+				indexers[api.URL] = &IndexTracker{
 					Name:       name,
-					Identifier: api,
+					Identifier: api.URL,
 					Source:     source,
-					Args:       args,
+					JsonPath:   api.JsonPath,
 				}
 			}
 			//now we definitely have one
```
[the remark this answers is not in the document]
hhio618: No problem, I'll take a look!

pkg/tracker/index_test.go
hhio618: Perfect, I felt like we need to make this code more testable. this seems like a good idea, I'm working on it.

pkg/tracker/index_test.go
```diff
@@ -3,40 +3,72 @@
...
+				},
+			},
+			Expected: []float64{454.534, 454.837},
+		},
+	}
+
+	// Test jsonpath parsing per test cases.
+	for _, testCase := range testCases {
+		payload, _ := testCase.IndexTracker.Source.Get()
+		vals, err := testCase.IndexTracker.parsePayload(payload)
```
krasi-georgiev, line 67:
```suggestion
		actual, err := testCase.IndexTracker.parsePayload(payload)
```
not a blocker just personal preference

krasi-georgiev:
Thanks, this is looking very good!  Just the failing tests. 

I think some tests are still flaky and it should be fixed in https://github.com/tellor-io/telliot/pull/317

PR #320, opened 2020-12-14
Cli replace with kong package and removed all global variables

JGcarv:


cmd/telliot/main.go
```diff
@@ -31,75 +31,83 @@ import (
...
+// var clt rpc.ETHClient
+
+// func ExitOnError(err error, operation string) {
+// 	if err != nil {
+// 		fmt.Fprintf(os.Stderr, "%s failed: %s\n", operation, err.Error())
+// 		cli.Exit(-1)
+// 	}
+// }
+
+func setup() (rpc.ETHClient, tellorCommon.Contract, tellorCommon.Account, error) {
```
krasi-georgiev, line 46: usually we do 
```suggestion
func setup() (rpc.ETHClient, *tellorCommon.Contract, *tellorCommon.Account, error) {
```
so that when there is an error can just 
```
return nil, nil, nil, errors.Wrap(err, "create rpc client instance")

File: cmd/telliot/main.go [hunk 2]
```diff
@@ -132,280 +140,300 @@ func AddDBToCtx(remote bool) error {
...
-`
+// const versionMessage = `
+//     The official Tellor cli tool %s (%s)
+//     -----------------------------------------
+// 	Website: https://tellor.io
+// 	Github:  https://github.com/tellor-io/telliot
+// `
 
-func App() *cli.Cli {
+func (m mineCmd) Run(logger log.Logger) error {
```
krasi-georgiev, line 153: I would keep the struct functions next to the struct type definition(`cmds.go`). this is idiomatic golang.

cmd/telliot/main.go
```diff
@@ -132,280 +140,300 @@ func AddDBToCtx(remote bool) error {
...
+	amount := EthereumInt{}
+	err = amount.Set(c.Amount)
+	if err != nil {
+		return errors.Wrapf(err, "parsing argument")
+	}
+	return ops.Approve(ctx, logger, client, contract, account, address.addr, amount.Int)
+}
+
+var cli struct {
+	LogConfig logConfigPath `type:"path" help:"path for log config file"`
```
krasi-georgiev, line 411: for paths to files and dirs best to use https://github.com/alecthomas/kong#custom-named-decoders

`existingfile`, `existingdir`

cmd/telliot/main.go
```diff
@@ -132,280 +140,300 @@ func AddDBToCtx(remote bool) error {
...
 	}
+	return nil
 }
 
-func newDisputeCmd(cmd *cli.Cmd) {
+func (n newDisputeCmd) Run(logger log.Logger, client rpc.ETHClient, contract tellorCommon.Contract, account tellorCommon.Account) error {
 	requestID := EthereumInt{}
+	err := requestID.Set(n.requestID)
+	if err != nil {
+		return errors.Wrapf(err, "parsing argument")
```
krasi-georgiev, line 325: It seems that instead of this kong has validation so all these checks can happen in the `Validate` function
https://github.com/alecthomas/kong#validation

`func (n newDisputeCmd) Validate(....`

JGcarv:
I don't know how to make the linter happy with all the Kong tags!

cmd/telliot/cmds.go
```diff
@@ -0,0 +1,316 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "setting up variables")
+	}
+
+	kong.BindTo(logger, (*log.Logger)(nil))
+	kong.BindTo(client, (*rpc.ETHClient)(nil))
+	kong.BindTo(ctx, (*context.Context)(nil))
+	kong.Bind(contract)
+	kong.Bind(account)
+	kong.Bind(cfg)
```
krasi-georgiev, line 52: This looks very suspiciously to a global var approach. Unless there is blocker I think we should just create these within the `Run(` function instead of this interesting `Bind` magic.

cmd/telliot/main.go
```diff
@@ -4,101 +4,28 @@
...
-	return nil
-}
-
 // migrateAndOpenDB migrates the tx costs and deletes the db.
 // The DB is always deleted because the price avarages calculations
 // is not calculated properly between restarts.
 // TODO don't do this and just improve the price calculations.
-func migrateAndOpenDB() (db.DB, error) {
-	cfg := config.GetConfig()
+func migrateAndOpenDB(cfg *config.Config) (db.DB, db.DataServerProxy, error) {
```
krasi-georgiev, line 24: We should keep functions to a single purpose. this this change the function no longer just migrates the db, but also creates a proxy. I don't see any reason why we can't just have another function `createDbProxy`. Long term having functions with small purpose allows to make components more composable, easyer to maintain, test and refactor.

cmd/telliot/cmds.go
```diff
@@ -0,0 +1,316 @@
...
+			}
+			// Start and wait for it to be ready.
+			if err := ds.Start(ctx); err != nil {
+				return errors.Wrapf(err, "starting data server")
+			}
+			<-ds.Ready()
+		}
+	}
+	// Start miner
+	DB := ctx.Value(tellorCommon.DataProxyKey).(db.DataServerProxy)
```
krasi-georgiev, line 263: Left over global context.

krasi-georgiev (commented):
Don't forget that this will also require a change log entry for any changes that are important for actual users - I think the main change is for the new config settings for the remote data server

krasi-georgiev:
> I don't know how to make the linter happy with all the Kong tags!

Please open an issue on the kong repo  with a link to this PR to see if they have some suggestions. As a last resort, we can add an ignore entry in the linter.

cmd/telliot/cmds.go
JGcarv: Hmm I agree it looks suspicious. I think there's two negatives of creating this in run functions:
1. It'll lead to a lot of repetition, every Run func will need to have this logic for parsing and executing the config file.
2. I think it'll be more verbose to write the logic to actually finding the config path, as I'm not sure if flags are easily accessible within run commands.

On the other hand, this global context is only used to pass arguments from setup to designated run functions, from there everything is explicitly passed through arguments, which is why I didn't bother too much.

But let me know what you think!

cmd/telliot/main.go
JGcarv: Sounds good!

cmd/telliot/main.go
```diff
@@ -110,331 +37,94 @@ func migrateAndOpenDB() (db.DB, error) {
...
 		}
 		dataProxy = proxy
 	}
-	database = DB
-	proxy = dataProxy
-	return nil
-}
-
-var GitTag string
-var GitHash string
```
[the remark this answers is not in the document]
JGcarv: I think there's a way indeed. I'll do a bit of research

cmd/telliot/cmds.go
krasi-georgiev: > It'll lead to a lot of repetition, every Run func will need to have this logic for parsing and executing the config file.

I think this is fine. We can extract the repeatable part in functions to reduce the repetition. Most golang devs(myself included) prefer to no magic whatsoever. I am not sure why the cli dev decided to use this global variables pattern.

> I think it'll be more verbose to write the logic to actually finding the config path, as I'm not sure if flags are easily accessible within run commands.

I don't understand the problem here. can you ping me and can discuss. 
This is what I imagine when you have a command that needs a config `mine --config-file=...` Inside run it should have access to the `config-file` variable right? This way it will just parse the config and build all the other required structs components etc.

cmd/telliot/main.go
```diff
@@ -110,331 +37,94 @@ func migrateAndOpenDB() (db.DB, error) {
...
 		}
 
+		return client, &contract, &account, nil
 	}
+	// Not sure why we need this case.
+	return nil, nil, nil, nil
+}
+
+var cli struct {
+	Config   configPath  `required type:"existingfile" help:"path to config file"`
```
krasi-georgiev, line 110: I think to be able to access the config value it needs to be added to every command instead of having it in the root.

krasi-georgiev: Look at this example. The arg `Paths` is accessible for the command `RmCmd` through `RmCmd.Paths`

```

type RmCmd struct {
  Force     bool `help:"Force removal."`
  Recursive bool `help:"Recursively remove files."`

  Paths []string `arg name:"path" help:"Paths to remove." type:"path"`
}

func (r *RmCmd) Run(ctx *Context) error {
  fmt.Println("rm", r.Paths)
  return nil
}
var cli struct {
  Rm RmCmd `cmd help:"Remove files."`
}

func main() {
  ctx := kong.Parse(&cli)
  // Call the Run() method of the selected parsed command.
  err := ctx.Run()
  ctx.FatalIfErrorf(err)
}
```

pkg/config/config.go
```diff
@@ -82,6 +82,7 @@ type Config struct {
 	ConfigFolder                 string                `json:"configFolder"`
 	LogLevel                     string                `json:"logLevel"`
 	Logger                       map[string]string     `json:"logger"`
+	RemoteMining                 bool                  `json:remoteMining`
```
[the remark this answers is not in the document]
JGcarv: Do you mean to have multiple config files or just command-specific fields in a single config file?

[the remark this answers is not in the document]
krasi-georgiev: We can keep a single file, but with some more structured format. Something like this would be a good start. But lets do it in a separate PR , for now just add `DataserverHost`, and `DataserverPort`

```
type Config struct{
	miner Mine type
	dataserver Dataserver type
	... other fields
}

{
	"contractAddress": "0xFe41Cb708CD98C5B20423433309E55b53F79134a",
    "publicAddress": "0x0e93f23278161bCa6F22D6268ca44691042ed437",
	"mine":{
		"host": "0.0.0.0", // It serves the metrics here. But later can expose some api.
    	"port": 5000,
		DataserverHost:... // Connect to this remote dataserver
		DataserverPort:...
		"dbFile": ".local/tellorDB", //  When DataserverHost is set this is not used.
	}
	"dataserver": {
		"host": "0.0.0.0",
    	"port": 5000,
		"dbFile": ".local/tellorDB",
	}
    
    "gasMax": 10,
    "configFolder": "configs",
    "profitThreshold": 50,
    "minConfidence": 0,
    "envFile": ".local/.env",
    "LogLevel": "debug",
    "Logger": {
        "config.Config": "INFO",
        "db.DB": "INFO",
        "rpc.client": "INFO",
        "rpc.ABICodec": "INFO",
        "rpc.mockClient": "INFO",
        "tracker.Top50Tracker": "INFO",
        "tracker.FetchDataTracker": "INFO",
        "pow.MiningWorker-0:": "INFO",
        "pow.MiningWorker-1:": "INFO",
        "pow.MiningTasker-0:": "INFO",
        "pow.MiningTasker-1:": "INFO",
        "tracker.PSRTracker": "INFO"
    }
}
```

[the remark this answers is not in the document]
JGcarv: Currently, ServerHost and ServerPort are only used for the remote connections, the local connection doesn't use those. I'll just leave them as is, ok?

cmd/telliot/cmds.go
```diff
@@ -0,0 +1,456 @@
...
+	case "withdraw":
+		return ops.WithdrawStake(ctx, logger, client, contract, account)
+	case "request":
+		return ops.RequestStakingWithdraw(ctx, logger, client, contract, account)
+	case "status":
+		return ops.ShowStatus(ctx, logger, client, contract, account)
+	default:
+		return errors.New("unknown stake command")
+	}
+}
```
krasi-georgiev, line 181: I think there should be a way to add subcommand without using string macthing.

krasi-georgiev (commented):
Great work!

Just few small nits and ready for merge.

We need an entry in the changelog that now the --config-file flag is after the command.

cmd/telliot/cmds.go
JGcarv: I think the only other way is to do it like dispute command, and each subcommand gets its own struct and such. Since these have the same requirements, I think make sense to group them together,but let me know if you think it's better the other way. Also, we can add to the help string the possible flags.

krasi-georgiev: Yeah these need to be consistent.

command --flags....

It is strange to see that some commands take the flags before the command like how it was for 

telliot --config  mine

Instead, all commands should be in the following format

telliot stake --config
telliot mine --config
telliot deposit --config

As long as this doesn't lead to a lot of duplicated code this makes it easier to remove/refactor/add commands.

JGcarv: Oh I see the problem. I will make the commands inconsistent indeed. I'll adjust that quickly

cmd/telliot/cmds.go
```diff
@@ -0,0 +1,512 @@
...
+	client, contract, account, err := createTellorVariables(ctx, cfg)
+	if err != nil {
+		return errors.Wrapf(err, "creating tellor variables")
+	}
+	// Create os kill sig listener.
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt)
+	exitChannels := make([]*chan os.Signal, 0)
+
+	// cfg := config.GetConfig()
```
krasi-georgiev, line 433:
```suggestion
```

pkg/config/config.go
```diff
@@ -96,6 +96,7 @@ type Config struct {
 	ConfigFolder                 string                `json:"configFolder"`
 	LogLevel                     string                `json:"logLevel"`
 	Logger                       map[string]string     `json:"logger"`
+	RemoteMining                 bool                  `json:"remoteMining"`
```
krasi-georgiev, line 99: you don't need this anymore. Now we have `RemoteDBHost string` when this is specified it should just do remote mining. When you change it also mention it in the changelog

pkg/util/logger.go
```diff
@@ -152,24 +152,22 @@ func xlateLevel(level LogLevel) logrus.Level {
...
+		lvl = level.AllowInfo()
+	case "debug":
+		lvl = level.AllowDebug()
+	default:
+		panic("unexpected log level")
 	}
+
+	logger := log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr))
+	logger = level.NewFilter(logger, lvl)
+	return log.With(logger, "ts", log.TimestampFormat(func() time.Time { return time.Now().UTC() }, "Jan 02 15:04:05.99 -0700"), "caller", log.DefaultCaller)
```
krasi-georgiev, line 172: Accidental revert of my change. I removed the timezon
```suggestion
	return log.With(logger, "ts", log.TimestampFormat(func() time.Time { return time.Now().UTC() }, "Jan 02 15:04:05.99"), "caller", log.DefaultCaller)
```

docs/changelog.md
```diff
@@ -10,6 +10,11 @@ We use _breaking :warning:_ to mark changes that are not backward compatible \(r
 
 ## Unreleased
 
+- [\#240](https://github.com/tellor-io/telliot/issues/240) Replaced the cli package to allow for command specific flags and configuration. Now all flags should be passed last. Example:
+  Instead of: `./telliot --config="config.json" stake deposit`, it becomes:`./telliot stake deposit --config="config.json"`
+
+- Removed the `RemoteMining` from the config. Remote is active when specifying a `RemoteDBHost` for the `Mine` command. See the `configs/config.json` for an example.
+
```
krasi-georgiev, line 17: gitbook doesn't like extra white spaces I think.
```suggestion
- Removed the `RemoteMining` from the config. Remote is active when specifying a `RemoteDBHost` for the `Mine` command. See the `configs/config.json` for an example.

```

docs/changelog.md
```diff
@@ -10,20 +10,24 @@ We use _breaking :warning:_ to mark changes that are not backward compatible \(r
 
 ## Unreleased
 
+* [\#240](https://github.com/tellor-io/telliot/issues/240) Replaced the cli package to allow for command specific flags and configuration. Now all flags should be passed last. Example:
+Instead of: `./telliot --config="config.json" stake deposit`, it becomes:`./telliot stake deposit --config="config.json"`
+
+* Removed the `RemoteMining` from the config. Remote is active when specifying a `RemoteDBHost` for the `Mine` command. See the `configs/config.json` for an example.
```
krasi-georgiev, line 16:
```suggestion
Removed the `RemoteMining` `-r` flag. Remote is active when specifying a `RemoteDBHost` for the `Mine` command. See the `configs/config.json` for an example.
```

docs/changelog.md
```diff
@@ -10,20 +10,24 @@ We use _breaking :warning:_ to mark changes that are not backward compatible \(r
...
+
 ### Changed
 
 ### Added
 
 ### Fixed
 
-## [v5.5.0](https://github.com/tellor-io/telliot/releases/tag/v5.5.0) - 2021.01.18
+## [v5.5.0](https://github.com/tellor-io/telliot/releases/tag/v5.5.0) * 2021.01.18
```
krasi-georgiev, line 24:
```suggestion
## [v5.5.0](https://github.com/tellor-io/telliot/releases/tag/v5.5.0) - 2021.01.18
```

docs/changelog.md
```diff
@@ -10,20 +10,24 @@ We use _breaking :warning:_ to mark changes that are not backward compatible \(r
...
 
 ### Changed
 
-* [\#372](https://github.com/tellor-io/telliot/pull/372) Split the configs of the mine and dataserver command to avoid confusion and be more explicit. This also fixes an issue where can't run a miner and dataserver on the same machine(now that the miner also runs an HTTP server to expose metrics). The config format has changed so users need to update their configs. See the `configs/config.jons` for an example of the new format.
+* [\#372](https://github.com/tellor*io/telliot/pull/372) Split the configs of the mine and dataserver command to avoid confusion and be more explicit. This also fixes an issue where can't run a miner and dataserver on the same machine(now that the miner also runs an HTTP server to expose metrics). The config format has changed so users need to update their configs. See the `configs/config.json` for an example of the new format.
```
krasi-georgiev, line 28:
```suggestion
* [\#372](https://github.com/tellor-io/telliot/pull/372) Split the configs of the mine and dataserver command to avoid confusion and be more explicit. This also fixes an issue where can't run a miner and dataserver on the same machine(now that the miner also runs an HTTP server to expose metrics). The config format has changed so users need to update their configs. See the `configs/config.json` for an example of the new format.
```

krasi-georgiev (approved):

PR #321, opened 2020-12-16
wip unifying the loggin configs with the main configs

JGcarv:


pkg/config/config.go
```diff
@@ -201,6 +211,24 @@ func validateConfig(cfg *Config) error {
 		}
 	}
 
+	var defaultEntries = []util.Entry{
```
JGcarv, line 214: Is this a good place to add the defaults?

I was in doubt between here and inside the setuoLoggingConfig function

krasi-georgiev: why not inside the config struct.
https://github.com/tellor-io/telliot/blob/master/pkg/config/config.go#L99

I think we should even rename it to `defaultConfig`

pkg/config/config.go
```diff
@@ -23,6 +23,11 @@ type Duration struct {
 	time.Duration
 }
 
+// type Entry struct {
```
krasi-georgiev, line 26: Left over?

pkg/config/config.go
```diff
@@ -136,35 +158,35 @@ func ParseConfig(path string) error {
 }
 
 func ParseConfigBytes(data []byte) error {
-	err := json.Unmarshal(data, &config)
+	err := json.Unmarshal(data, &defaultConfig)
```
krasi-georgiev, line 161: hm, now I see why it was called `config` and not `defaultconfig` it gets confusing when we use it here. Can we change this to make it more explicit that the default confit is parsed into the final config to use. Something like:
```suggestion
	config := &defaultConfig;
	err := json.Unmarshal(data, config)
```
This doesn't change anything, but later it is more clear when we have something like:`config.PublicAddress` , instead of `defaultConfig.PublicAddress` - this one looks like we still use the default config which is misleading.

pkg/util/logConfig.go
```diff
@@ -3,15 +3,7 @@
...
-	"encoding/json"
-	"fmt"
-	"os"
-
-	"github.com/pkg/errors"
-)
-
-// Entry holds specific component log level.
+// // Entry holds specific component log level.
 type Entry struct {
```
krasi-georgiev, line 7: do we still need this struct?

cmd/telliot/main.go
```diff
@@ -222,22 +227,20 @@ func balanceCmd(cmd *cli.Cmd) {
 	cmd.VarArg("ADDRESS", &addr, "ethereum public address")
 	cmd.Spec = "[ADDRESS]"
 	cmd.Action = func() {
-		// Using values from context, until we have a function that setups the client and returns as values, not as part of the context
-		commonAddress := ctx.Value(tellorCommon.PublicAddress).(common.Address)
 		var zero [20]byte
 		if bytes.Equal(addr.addr.Bytes(), zero[:]) {
-			addr.addr = commonAddress
+			addr.addr = acc.Address
```
krasi-georgiev, line 232: I really hope we can get rid of these global variables with the cli refactoring. the final goal is to use only variables that are passed to a function. Just mentioning it here to make sure we are on the same page.

JGcarv: Yes, they'll be gone!

krasi-georgiev (approved):

PR #325, opened 2020-12-21
Don't resubmit when the miner has already submitted for the current oracle block

### LINKED ISSUE #294: Add a check to NOT submit a solution when the miner has already submitted for the current block.
A miner can't submit more than once for an oracle mining block so more than one transaction for a given mining block causes failed transactions and  wasted gas.

It should be possible to check the contract  if the miner already submitted for the current block and only then submit.

themandalore:
fixes #294

### DIFF
--- a/pkg/ops/miningManager.go
+++ b/pkg/ops/miningManager.go
@@ -53,6 +53,7 @@ type MiningMgr struct {
 	tasker          WorkSource
 	solHandler      SolutionSink
 	solutionPending *pow.Result
+	submitted       bool
 	database        db.DataServerProxy
 	contractGetter  *getter.TellorGetters
 	cfg             *config.Config
@@ -93,6 +94,7 @@ func CreateMiningManager(
 		tasker:          nil,
 		solutionPending: nil,
 		solHandler:      nil,
+		submitted:		 false,
 		contractGetter:  getter,
 		cfg:             cfg,
 		database:        database,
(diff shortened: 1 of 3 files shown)


themandalore:
fixes #294

krasi-georgiev:
> fixes #294

this only works from the first comment.

krasi-georgiev:
This solution has race condition as it reads the variable while writing to it in a goroutine. Can add locking, but I think a more simple solution would be to just do error matching and ignore an error 
 with the text `Miner already submitted the value`


```
tx, err := mgr.solHandler.Submit(ctx, solution)
if err != nil {
		level.Error(mgr.logger).Log("msg", "submiting a solution", err)
		mgr.submitFailCount.Inc()
		// No need to resubmit when already submitted for the current oracle block.
		if err.Error() == "Miner already submitted the value"{
			mgr.solutionPending = nil
		}
		continue
}
```

themandalore:
> tx, err := mgr.solHandler.Submit(ctx, solution)
> if err != nil {
> 		level.Error(mgr.logger).Log("msg", "submiting a solution", err)
> 		mgr.submitFailCount.Inc()
> 		// No need to resubmit when already submitted for the current oracle block.
> 		if err.Error() == "Miner already submitted the value"{
> 			mgr.solutionPending = nil
> 		}
> 		continue
> }

Doesn't this not prevent it though?  Or are you saying I leave my changes and then add this to the code too?  I think if the error is that you already sumbitted the value, why would you put the solutionPending to nil?

PR #342, opened 2020-12-31
Make the minimum submit configurable

krasi-georgiev:
This is so that in development can use TellorPlayground to submit
transactions faster which will speed up the debugging during
development. At the moment it doesn't work because of the contract
binding, but at some point will see how we can fix this.

### DIFF
--- a/configs/config.json
+++ b/configs/config.json
@@ -6,6 +6,20 @@
     "gasMax": 10,
     "dbFile": "/tellorDB",
     "configFolder": "/configs",
+    "envFile": "configs/.env.example",
     "logLevel": "info",
-    "envFile": "configs/.env.example"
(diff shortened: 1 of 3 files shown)


themandalore:
so how would this work with the playground?  Could you just set the contract to be a playground address and it would auto submit?  I think the problem is that the playground uses a different function for submitting, so we'd need to write an interface

krasi-georgiev:
yes I haven't figured out the details yet but just shooting ideas on how to improve the local development instead of waiting 15min for every submit. Regardless of how we implement the calls to the yellow playground would need this period to be configurable in telliot that is why I thought to just add now.

themandalore (approved):

PR #347, opened 2021-01-02
On-chain index trackers for Uniswap, Balancer smart contracts

### LINKED ISSUE #280: Onchain contract calls in the tracker package / indexes.json
We should have a way to grab on-chain events for transformation.  

E.G.: Ampleforth price on Uniswap and volume - then do the daily VWAP or TWAP

hhio618:
### Description
What's done:
- Balancer on-chain price.
- Uniswap on-chain price.
- Update `index.go` to handle UniswapGetter, BalancerGetter index types.
- Mock on-chain calls, unittest.

### Refs
closes #280


krasi-georgiev (commented):
I just made this simple tool that can be used to generate the bindings for this Pr and simplify it. 
https://github.com/krasi-georgiev/telliot/blob/go-bindings/scripts/bindings/bindings.go (as part of this PR https://github.com/tellor-io/telliot/pull/356)
When merged should give it a try and adjust if needed to be able to generate the bindings for this PR as well.

And don't forget to add an entry in the changelog as this a noise new feature that the users should be notified about :)

pkg/tracker/uniswapTWAP.go
```diff
@@ -0,0 +1,232 @@
...
+
+	tellorCommon "github.com/tellor-io/telliot/pkg/common"
+	"github.com/tellor-io/telliot/pkg/db"
+	"github.com/tellor-io/telliot/pkg/rpc"
+	uniswapcontract "github.com/tellor-io/telliot/pkg/tracker/uniswap"
+)
+
+// UniswapPriceCumulativeLast will be used to save/load data into/from the DB, so that
+// we can minimize onchain contract calls.
+type UniswapPriceCumulativeLast struct {
```
krasi-georgiev, line 28:
```suggestion
type UniswapPriceCumulative struct {
```
I think this is infered.

pkg/tracker/uniswapTWAP.go
```diff
@@ -0,0 +1,232 @@
...
+	Price0CumulativeLast string
```
krasi-georgiev, line 29:
```suggestion
	PriceCumulative string
```

pkg/tracker/uniswapTWAP.go
```diff
@@ -0,0 +1,232 @@
...
+	Decimals             uint8
+	Decimals0            uint8
+	Decimals1            uint8
+	TimestampLast        uint32
+	CalculatedPrice0Last float64
```
krasi-georgiev, line 34:
```suggestion
	PriceTWAP float64
```

pkg/tracker/uniswapTWAP.go
```diff
@@ -0,0 +1,232 @@
...
+)
+
+// UniswapPriceCumulativeLast will be used to save/load data into/from the DB, so that
+// we can minimize onchain contract calls.
+type UniswapPriceCumulativeLast struct {
+	Price0CumulativeLast string
+	Decimals             uint8
+	Decimals0            uint8
+	Decimals1            uint8
+	TimestampLast        uint32
```
krasi-georgiev, line 33:
```suggestion
	Timestamp        uint32
```

pkg/tracker/uniswapTWAP.go
```diff
@@ -0,0 +1,232 @@
...
+	DB := ctx.Value(tellorCommon.DBContextKey).(db.DB)
+
+	// getting price cumulative last onchain.
+	priceCumulativeLast, err := u.getCumulativePriceLast()
+	if err != nil {
+		return nil, errors.Wrap(err, "getting price cumulative last")
+	}
+	// Try to get previous price0Cumulative, timestamp from DB so that we minimize contract calls.
+	var priceCumulative *UniswapPriceCumulativeLast
+	priceCumulative, err = u.pullPriceFromDB(DB)
```
krasi-georgiev, line 75:
```suggestion
	priceCache, err = u.pullPriceFromDB(DB)
```

Later this will make it more clear to understand the logic like
`price.Timestamp == priceCache.Timestamp`

pkg/tracker/uniswapTWAP.go
```diff
@@ -0,0 +1,232 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "getting price cumulative from DB")
+	}
+	// Return the price data based on conditions.
+	var priceData []float64
+	if priceCumulative == nil {
+		priceData = []float64{0}
+
+	} else if priceCumulative.Price0CumulativeLast == priceCumulativeLast.Price0CumulativeLast {
```
krasi-georgiev, line 84: Why not just compare the timestamp to make it more explicit what is happening here?
```suggestion
	} else if price.Timestamp == priceCache.Timestamp {
```

pkg/tracker/uniswapTWAP.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+	// Try to get previous price0Cumulative, timestamp from DB so that we minimize contract calls.
+	var priceCumulative *UniswapPriceCumulativeLast
+	priceCumulative, err = u.pullPriceFromDB(DB)
+	if err != nil {
+		return nil, errors.Wrap(err, "getting price cumulative from DB")
+	}
+	// Return the price data based on conditions.
+	var priceData []float64
+	if priceCumulative == nil {
```
krasi-georgiev, line 81: This doesn't look right - when the db doesn't return any price it should fall back to get it from uniswap right?

pkg/tracker/uniswapTWAP.go
```diff
@@ -0,0 +1,232 @@
...
+		priceData = []float64{0}
```
krasi-georgiev, line 82: This doesn't look right - when the db doesn't return any price it should fall back to get it from uniswap right?

pkg/tracker/balancer.go
```diff
@@ -0,0 +1,137 @@
...
+	// use balancer pool own GetSpotPrice to minimize onchain calls.
+	price, err := b.getSpotPrice(pair)
+	if err != nil {
+		return nil, errors.Wrap(err, "getting price info from balancer pool")
+	}
+	// Output to index tracker.
+	return json.Marshal([]float64{price})
+}
+
+func (b *BalancerGetter) getPair() (pair *BalancerPair, err error) {
```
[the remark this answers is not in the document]
hhio618: Very good point!

krasi-georgiev (commented):
.

pkg/tracker/uniswapTWAP.go
hhio618: Yeah, agree!

pkg/tracker/uniswapTWAP.go
hhio618: This is needed for calculating a TWAP price according to Uniswap docs. 
In the time when there is no previous price, we can either return zero and calculate the price on the next call or look back into previous blocks and pick another block for our calculations! 

Which one do you think is best?

pkg/tracker/uniswap.go
```diff
@@ -0,0 +1,168 @@
...
+
+func (u *Uniswap) getSpotPrice() (*big.Float, error) {
+	var pairContract *uniswap.IUniswapV2PairCaller
+	var err error
+	pairContract, err = uniswap.NewIUniswapV2PairCaller(common.HexToAddress(u.address), u.client)
+	if err != nil {
+		return nil, errors.Wrap(err, "getting pair contract")
+	}
+
+	// ensure that there's liquidity in the pair.
```
krasi-georgiev, line 66:
```suggestion
	// Ensure that there's liquidity in the pair.
```

pkg/rpc/mock.go
```diff
@@ -45,6 +57,16 @@ type CurrentChallenge struct {
 	Tip           *big.Int
 }
 
+// CurrentReserves holds details about the current reserves on the uniswap pair contract.
+// Note: Each Uniswap smart contract, or pair, manages a liquidity pool made up of reserves of two ERC-20 tokens.
```
krasi-georgiev, line 61: Reserve here is how much liquidity is in the pool? If yes please add this extra details.

scripts/bindings/bindings.go
```diff
@@ -62,17 +80,36 @@ func generate(address string, contractFPath string) {
 		log.Fatal(err)
 	}
 
-	if err := ioutil.WriteFile(contractFPath, []byte(src[0].SourceCode), os.ModePerm); err != nil {
+	var contractFiles = []string{contractFPath}
```
krasi-georgiev, line 83: Let's extract the solc detection and  download into a separate func to make this one shorter and keep functions with single-purpose scope.

pkg/tracker/ampl_test.go
```diff
@@ -41,6 +41,10 @@ func TestAmpl(t *testing.T) {
 	indexers = append(indexers, amplBtcTrackers...)
 	for i := 0; i < 288; i++ {
 		for _, indexer := range indexers {
+			// Ignore on-chain trackers, as they could be tested in other test cases.
+			if strings.Contains(indexer.Identifier, "ethereum:") {
```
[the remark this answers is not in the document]
hhio618: I can fix this by adding a Type field to the IndexTracker struct.

scripts/bindings/bindings.go
hhio618: Done!

pkg/tracker/index.go
```diff
@@ -160,27 +172,27 @@ type IndexTracker struct {
 }
 
 type DataSource interface {
-	Get() ([]byte, error)
+	Get(ctx context.Context) ([]byte, error)
```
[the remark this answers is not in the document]
krasi-georgiev: On a second thought I now see that you use the ctx to get a client instance but we removed this pattern and now don't use the context for storing these so this param can be removed.

```
u.client = ctx.Value(tellorCommon.ClientContextKey).(rpc.ETHClient)
```
Instead of this need to pass it in the `NewUniswap` constructor. See the pr for removing the global context.

pkg/tracker/ampl_test.go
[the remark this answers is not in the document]
krasi-georgiev: Thanks!

krasi-georgiev (approved):

PR #359, opened 2021-01-06
Refactored the solution sibmit logic

krasi-georgiev:
fixes: #325 #294

A simplified version that should be easier to follow and trying to fix the linked issued.

I will test it on the server and will let you know if all looks ok.


krasi-georgiev:
Ok after some more testing the problem is the following. 


There are 2 restrictions on the oracle 
 - a miner can submit once every 15min
 - a miner can submit only once for a current block

So when there are only 4 miners the following happens
 1. a miner submits solution 
 1. 15min are passing but no new block is started 
 1. miner calls `getNewCurrentVariables` this returns a new challenge 
 1. miner submits again 
 1. transaction fails because a miner can submit only once for the same oracle block.
 2. the whole loop repeats again which leads to many failed transactions and wasted gas.

I am not sure why `getNewCurrentVariables` returns a new challenge on every call - probably because it uses time to generate the challenge. 

I think there are few options to solve this problem
 - `getNewCurrentVariables` should also return the current oracle block Id - the miner can match that has already provided a solution for the current block.
 - query the oracle if the miner has already submitted for the current block - 5 chain requests will be needed
 - add a helper function to `lens` - minerCanSubmit to combine the 15min time check and whether the miner has submitted for the current block. This will reduce the calls to 1. The only problem here is that in the `env` file need to add the lens chain address.

themandalore:
> Ok after some more testing the problem is the following.
> 
> There are 2 restrictions on the oracle
> 
> * a miner can submit once every 15min
> * a miner can submit only once for a current block
> 
> So when there are only 4 miners the following happens
> 
> 1. a miner submits solution
> 2. 15min are passing but no new block is started - to start a new block we need 5 miners - 5 solutions.
> 3. miner calls `getNewCurrentVariables` this returns a new challenge
> 4. miner submits again
> 5. transaction fails because a miner can submit only once for the same oracle block.
> 6. the whole loop repeats again which leads to many failed transactions and wasted gas.
> 
> I am not sure why `getNewCurrentVariables` returns a new challenge on every call - probably because it uses time to generate the challenge.
> 
> I think there are few options to solve this problem
> 
> * `getNewCurrentVariables` should also return the current oracle block Id - the miner can match that has already provided a solution for the current block.
> * query the oracle if the miner has already submitted for the current block - 5 chain requests will be needed
> * add a helper function to `lens` - minerCanSubmit to combine the 15min time check and whether the miner has submitted for the current block. This will reduce the calls to 1. The only problem here is that in the `env` file need to add the lens chain address.

getNewCurrentVariables shouldn't return a different challenge...this is a bug if it does.  AFAIK , what happens is that it currently checks for a newChallenge...but in the process of doing that, it realizes it's been 15 minutes.  The autosubmit in the code overrides the fact that it's an old challenge (as it should, we've been mining it for 15 minutes), and submits.  The problem is we don't currently check whether we've submitted.   
and what was wrong with my solution of just storing a variable if you submitted for a given challenge? It prevents on-chain calls and solves the issue

krasi-georgiev:
Ok I just double-checked and you are right `currentChallenge` doesn't change for the same block. The reason why I am so stubborn about adding this new variable is that I don't see how it fixes the issue and OTOH the variable `mgr.solutionPending` has exactly the purpose to track when a solution has been sent and to never resend. It seems like I am missing something here so will spend a bit more time on this.

krasi-georgiev:
With this refactoring mining will use less CPU as well.

pkg/ops/miningManager.go
```diff
@@ -152,102 +156,140 @@ func (mgr *MiningMgr) Start(ctx context.Context) {
...
+		if work == nil {
+			return
+		}
+		var ids []int64
+		for _, id := range work.Challenge.RequestIDs {
+			ids = append(ids, id.Int64())
+		}
+		// For oracle blocks that are more than 15mins old can submit any solution.
+		// No need to send it for minging.
+		if mgr.isInstantSubmit() {
```
themandalore, line 240: Now this only get's called when a challenge is new (it returns on line 231)...
You want it to check for instantSubmit every trackerCycle

krasi-georgiev:
this will be fixed properly in https://github.com/tellor-io/telliot/pull/399

PR #363, opened 2021-01-08
IndexTracker custom cycle times

### LINKED ISSUE #357: Refactor the trackers to allow interval pull per URL
Will leave a global interval cycle for all URLs, but will also allow setting custom interval per URL.

This way can adjust the poll interval for  APIs that are throttled.

Related to https://github.com/tellor-io/telliot/pull/355

themandalore:
closes #357

### DIFF
--- a/configs/indexes.json
+++ b/configs/indexes.json
@@ -157,6 +157,11 @@
         {
             "URL": "https://api.defillama.com/charts",
             "param": "$[-1:].totalLiquidityUSD"
+        },
+        {
+            "URL": "https://data-api.defipulse.com/api/v1/defipulse/api/MarketData",
+            "param": "$.All.total",
+            "interval": 600
         }
     ],
     "EOS/USD": [
--- a/pkg/tracker/index.go
+++ b/pkg/tracker/index.go
@@ -14,6 +14,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/benbjohnson/clock"
 	"github.com/joho/godotenv"
@@ -102,6 +103,15 @@ func parseIndexFile(cfg *config.Config, DB db.DB) (trackersPerURL map[string]*In
 				default:
 					return nil, nil, errors.New("unknown index type for index object")
 				}
+
+				apiInterval := cfg.TrackerSleepCycle.Duration
+				if api.Interval.Duration > 0 {
+					apiInterval = api.Interval.Duration
+				}
+				if apiInterval < cfg.TrackerSleepCycle.Duration {
+					return nil, nil, errors.New("api interval can't be smaller than the global tracker cycle")
+				}
+
 				// Default value for the parser.
 				if api.Parser == "" {
 					api.Parser = jsonPathIndexParser
@@ -111,6 +121,7 @@ func parseIndexFile(cfg *config.Config, DB db.DB) (trackersPerURL map[string]*In
 					Identifier: api.URL,
 					Source:     source,
 					DB:         DB,
+					Interval:   apiInterval,
 					Param:      api.Param,
 				}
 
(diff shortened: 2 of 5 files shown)


pkg/tracker/index.go
```diff
@@ -151,14 +156,17 @@ func BuildIndexTrackers(cfg *config.Config, db db.DB) ([]Tracker, error) {
...
 }
+
 type IndexTracker struct {
 	DB         db.DB
 	Name       string
 	Identifier string
 	Symbols    []string
 	Source     DataSource
 	JSONPath   string
+	Interval   uint //seconds minimum that we wait before pulling again
```
krasi-georgiev, line 169: Please follow Golang commentary guidelines.

Here and in few other places

pkg/tracker/index.go
```diff
@@ -182,25 +190,36 @@ func (j *JSONfile) Get() ([]byte, error) {
...
 func (i *IndexTracker) Exec(ctx context.Context) error {
-	payload, err := i.Source.Get()
-	if err != nil {
-		return err
+	now := clck.Now()
+	currentVal := apiOracle.GetLatest(i.Identifier)
+	runID := true
+	if currentVal != nil {
+		if uint(now.Sub(currentVal.Created)) < i.Interval {
+			runID = false
```
krasi-georgiev, line 198:
```suggestion
			return nil
```
Instead of this, you can just return here.

Btw I think a more simple approach could be to just add a variable `lastRunTimestamp`

This way you can have something like
```
if now.Sub(i.lastRunTimestamp)< i.Interval{
	return nil
}
i.lastRunTimestamp = now

```

configs/indexes.json
```diff
@@ -156,7 +156,8 @@
     "DEFITVL": [
         {
             "URL": "https://api.defillama.com/charts",
-            "jsonPath": "$[-1:].totalLiquidityUSD"
+            "jsonPath": "$[-1:].totalLiquidityUSD",
+            "interval":600
```
krasi-georgiev, line 160: I think defilama doesn't need an interval. Need to add the defipulse here with the interval.

```
{
	"URL": "https://data-api.defipulse.com/api/v1/defipulse/api/MarketData",
    "jsonPath": "$.All.total",
	"interval":600
   }
```

pkg/tracker/index.go
```diff
@@ -84,12 +84,17 @@ func parseIndexFile(cfg *config.Config, DB db.DB) (trackersPerURL map[string]*In
 					source = &JSONfile{filepath: filepath.Join(cfg.ConfigFolder, api.URL)}
 					name = filepath.Base(api.URL)
 				}
+				minimumInterval := uint(cfg.TrackerSleepCycle.Duration) / 2
```
krasi-georgiev, line 87: Why is this divided by 2?

krasi-georgiev:
Also I just remembered that there should be some check that no index interval should be smaller than the global tracker cycler.

pkg/tracker/index.go
themandalore: I was debating this...say the trackerCycle is 10 seconds.  If we check 9 seconds later, should it update if no interval specified?

themandalore: yeah, I'm deleting it, I think keep it flat and query these things less

themandalore:
added

pkg/tracker/index.go
krasi-georgiev: What do you think about this suggestion? I find it a bit easier to understand

pkg/tracker/index.go
krasi-georgiev: But we should deff have some check that the custom API interval is the same or smaller than the global  interval

krasi-georgiev (approved):

PR #365, opened 2021-01-10
add TestIndexParsable

hhio618:
### Description
This will add a test that checks if indexes.json index trackers are parsable.


krasi-georgiev:
The test is failing because one API needs a key. I have added it to the repo as a secret so you just need to add it as an env var in the github test workflow

krasi-georgiev:
Need to do the same for couverage test.

krasi-georgiev (approved):

PR #366, opened 2021-01-10
Structured index parsing to allow different parsers

### LINKED ISSUE #358: Refactor the index file format to make the fields more explicit
Current:
```
"URL": "https://api.coingecko.com/api/v3/simple/price?ids=ampleforth&vs_currencies=usd",
"jsonPath": "$.ampleforth.usd"
```

```
URL:... where to fetch it from
type: - how to fetch
	http
	etherium
	file
format: - how to parse
	jsonPath
```

this will avoid using string matching like in https://github.com/tellor-io/telliot/blob/master/pkg/tracker/index.go#L76 - any string matching is error prone

Also with this the Exec interface becomes like - fetch -> format and give much more flexibility on sources and formats

hhio618:
### Refs
closes #358

### DIFF
--- a/configs/indexes.json
+++ b/configs/indexes.json
@@ -1,662 +1,480 @@
 {
-  "ALGO/USD": [
-    {
-      "URL": "https://api.binance.com/api/v1/klines?symbol=ALGOUSDT&interval=1d&limit=1",
-      "type": "http",
-      "format": "jsonPath",
-      "jsonPath": "$[0][4]"
-    }
-  ],
-  "AMPL/BTC": [
-    {
-      "URL": "https://api-pub.bitfinex.com/v2/tickers?symbols=tAMPBTC",
-      "type": "http",
-      "format": "jsonPath",
-      "jsonPath": "$[0][7,8]"
-    }
-  ],
-  "AMPL/USD": [
-    {
-      "URL": "https://api.coingecko.com/api/v3/simple/price?ids=ampleforth&vs_currencies=usd",
-      "type": "http",
-      "format": "jsonPath",
-      "jsonPath": "$.ampleforth.usd"
-    },
-    {
-      "URL": "https://api-pub.bitfinex.com/v2/tickers?symbols=tAMPUST",
-      "type": "http",
-      "format": "jsonPath",
-      "jsonPath": "$[0][7,8]"
-    },
-    {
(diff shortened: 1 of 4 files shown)


configs/indexes.json
```diff
@@ -1,480 +1,662 @@
...
+      "format": "jsonPath",
+      "jsonPath": "$[0][4]"
+    }
+  ],
+  "ZRX/USD": [
+    {
+      "URL": "https://api.binance.com/api/v1/klines?symbol=ZRXUSDT&interval=1d&limit=1",
+      "type": "http",
+      "format": "jsonPath",
+      "jsonPath": "$[0][4]"
```
krasi-georgiev, line 659: Now that I see it I think these are better names.
```suggestion
      "parser": "jsonPath"
      "param": "$[0][4]"
```

The same field will be used for all parsers so it needs to be more generic.
for example, imagine the following combinations
```
parser: XML
param: x.v.b

parser: jsonPath
param: $[0][4]
```

So essentially this becomes the parser parameter.

krasi-georgiev (commented):
This will require an entry in the change log as users would need to update the index file format with this change.

Also I think we can make `http` and `jsonPath` the defaults to reduce the verbosity in the index.json file.
So essentially run `json.Unmarshal(byteValue, &baseIndexes)` and then run a loop `if baseIndexes[symbol].type=="" { baseIndexes[symbol].type="http" }` and the same for `jsonPath`

docs/changelog.md
```diff
@@ -12,6 +12,7 @@ We use _breaking :warning:_ to mark changes that are not backward compatible \(r
 
 #### Changed
 * [\#321](https://github.com/tellor-io/telliot/pull/321) Unified all configuration files. LoggingConfig and LogLevel now reside in the main config file.
+* [\#366](https://github.com/tellor-io/telliot/pull/366) Refactor `indexes.json` file (Renamed "jsonPath" -> "param").
```
krasi-georgiev, line 15: This needs a bit better description. 
```suggestion
* [\#366](https://github.com/tellor-io/telliot/pull/366) Refactored the `index.json` parsing to be more flexible and allow using different parsers. With the notion of `parser` and `param` can allow combining different parsers and parsers parameters. The default is still `jsonPath`, but current users need to rename 	jsonPath` to `param` in their `index.json` file.
```

krasi-georgiev (approved):

PR #377, opened 2021-01-20
Fixing null pointer exception

JGcarv:
The default string for remoteDB host was making the checks to identify weather we're in remote mining mode fail.


krasi-georgiev:
superseded by https://github.com/tellor-io/telliot/pull/382

PR #378, opened 2021-01-20
Gpu removal

### LINKED ISSUE #267: Remove the GPU mining code as noone uses it.
this should be done only after fixing the stratum pool mining in https://github.com/tellor-io/TellorMiner/issues/266.

This will unblock is to have automated building and testing for windows and osx.

JGcarv:
this closes  #267

I only need some guidance on what to do with the dockerfile, as there's some commands specific for gpu mining dependancy


krasi-georgiev:
These can be removed
https://github.com/tellor-io/telliot/blob/master/Makefile#L59-L60
https://github.com/tellor-io/telliot/blob/master/Makefile#L73
https://github.com/tellor-io/telliot/blob/master/Makefile#L91

Delete this file
pkg/pow/kernelSource.go

Remove these lines
https://github.com/tellor-io/telliot/blob/master/Dockerfile#L4

Replace this
https://github.com/tellor-io/telliot/blob/master/Dockerfile#L12
with
`RUN apt-get install -y --no-install-recommends ca-certificates && rm -rf /var/lib/apt/lists/*

Makefile
```diff
@@ -118,7 +114,7 @@ lint: go-lint shell-lint
 # to debug big allocations during linting.
 .PHONY: go-lint
 go-lint: check-git deps $(GOLANGCI_LINT) $(FAILLINT)
-go-lint: pkg/pow/kernelSource.go
+go-lint:
```
krasi-georgiev, line 117:
```suggestion
```

krasi-georgiev (commented):
Great! I just tested it and now the windows build works fine so I would say lets add it to the automatic build for the releases. The release name now is only telliot so we need to change it to telliot-linux and telliot-windows.exe

We need to update https://github.com/tellor-io/telliot/blob/master/.github/workflows/release.yml for this. Ping if need any help.

krasi-georgiev:
Actually let's make the windows build in a separate PR, will open an issue for it.

krasi-georgiev (approved):

PR #383, opened 2021-01-22
Auto detect network for blockchain trackers

### LINKED ISSUE #381: When using oracle trackers need to detect the network and get the correct contract address
The address for Uniswap and Balancer in the current index.json is pointing to the contracts in mainnet so when running the cli for rinkeby it returns errors as the contract addresses there are different.

No idea how to fix it nicely.

hhio618:
### Refs
resolves #381

### DIFF
--- a/configs/indexes.json
+++ b/configs/indexes.json
@@ -31,14 +31,14 @@
     ],
     "AMPL/USDC": [
         {
-            "URL": "ethereum:0x7860e28ebfb8ae052bfe279c07ac5d94c9cd2937",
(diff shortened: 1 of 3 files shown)


krasi-georgiev (commented):
On a second thought I think the PR will become too big so lets keep this PR only for the uniswap/balancer and open a new one for Tellor contract addresses.

For Tellor instead of this approach can just hard code it in
pkg/contracts/contracts.go
With this change can remove all references in the config for the Contract address(do a global search) and also need to look through all docs to check what needs updating there as well.


Can you also delete here or in a separate PR
pkg/common/contextKeys.go 
We shouldn't be using these anymore.

krasi-georgiev (approved):

PR #386, opened 2021-01-25
Hard-code Tellor contract address and remove it from the config

hhio618:
### Description
This will remove `ContractAddress` from the config file and hard-code it in the `contracts` package.


krasi-georgiev (commented):
Can you remove it from the example configs and the test configs.

Also look through the docs to see if anything needs adjusting.

krasi-georgiev (approved):

PR #388, opened 2021-01-26
Increase interval for cryptocompare to be below the threshold

krasi-georgiev:
After running it for a while it returns
```
 {"Response":"Error","Message":"You are over your rate limit please upgrade your account!","HasWarning":false,"Type":99,"RateLimit":{"calls_made":{"second":1,"minute":11,"hour":485,"day":10973,"month":82758,"total_calls":82758},"max_calls":{"second":20,"minute":300,"hour":3000,"day":7500,"month":50000}},"Data":{}}
```

With this change it should be below the threshold


JGcarv:
Can we add the increase in coin market cap in this PR as well?

krasi-georgiev:
yeah good idea, lets test all API's a bit longer and when all stable will adjust all in the same PR

JGcarv (approved):

PR #389, opened 2021-01-26
Make all logging consistent and now only using the log kit pkg

TheRocketCat:
fixes: https://github.com/tellor-io/telliot/issues/185

When changing from fmt.Pr and Logrus to the go-kit logger i tried to follow the flow (with minimal code change), in terms of editing comments and making code changes.

There was one fmt.Pr i did not change, it was in the cmd/telliot/cmd.go Run function (i think), because it seemed to actually want to print info about the programs version.

Also by the end of this i had an extra file, that i have no idea if it actually is part of the project, it seemed to appear after i pulled some changes to make sure my code was auto mergable. 

I did not remove it (even tho i cant find it in your repo), because when i realised it seemed entierly new i had already editeded its fmt.Pr* and did not have the heart to potentially discard my extra work. File in question pkg/db/localDataProxy.go


krasi-georgiev (commented):
Thanks, it is starting to look much better :+1: 

There was a little bit of oversight on my end. We want to have proper logging only for the mine and dataserver command as these are the only long running ones. For the other cmds that just execute an action and exit would be better to keep the use of `fmt.Print` as this will present the result in a cleaner way than with a log line.


With this PR we want to also remove `cfg.LogLevel` from the config.
so instead of `createLogger` we will have 2 functions `NewLogger` and `ApplyFilter`


```
func NewLogger() log.Logger{
	logger := log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr))
	return log.With(logger, "ts", log.TimestampFormat(func() time.Time { return time.Now().UTC() }, "Jan 02 15:04:05.99"), "caller", log.DefaultCaller)
}

func ApplyFilter(logLevel string) (log.Logger,error) {
	lvl := level.AllowInfo()
	if level,ok:= cfg.Logger[ComponentName];ok{
		switch level {
		  case "error":
			  lvl = level.AllowError()
		  case "warn":
			  lvl = level.AllowWarn()
		  case "info":
			  lvl = level.AllowInfo()
		  case "debug":
			  lvl = level.AllowDebug()
		  default:
			  return nil,errors.Errorf("unexpected log level:%v",level)
		  }
	}

	return level.NewFilter(logger, lvl),nil
}
```

basically each component in the cli has its own log level set by the config file.

In each major component constructor that takes a logger as a parameter should have something like:

```
const ComponentName = "trackers"
New(logger log.Logger, ctx...., cfg){
	logger,err = ApplyFilter(cfg,ComponentName,logger)
}
```

This granular setup is important only for the dataserver and mine command as these are the only long-running commands, all the others just execute a single action and exit.

tl-dr -  all commands should use `NewLogger` in their `Run()` 
Additionaly mine and dataserver cmds should also  use `ApplyFilter` in the constructors of some internal components where the logging might need more granular filtering.


At the end the logrus import should be removed "github.com/sirupsen/logrus"

pkg/apiOracle/valueOracle.go
```diff
@@ -11,12 +11,14 @@ import (
...
 
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
 	"github.com/pkg/errors"
 	"github.com/tellor-io/telliot/pkg/config"
 	"github.com/tellor-io/telliot/pkg/util"
 )
 
-var logger = util.NewLogger("apiOracle", "valueOracle")
+var logger = log.With(util.NewLogger(), "apiOracle", "valueOracle")
```
krasi-georgiev, line 21: There should be no global logger variables, these should be passed in the component constructors.

also lets extract the component names into constants so that can use these with the log configuration for the level
```suggestion

const ComponentName = "apiOracle"
```

pkg/dataServer/dataServer.go
```diff
@@ -43,6 +44,11 @@ func CreateServer(
 	if err != nil {
 		return nil, errors.Wrapf(err, "creating data server tracker runner instance")
 	}
+
+	filterLogger, err := util.ApplyFilter(*config, "DataServer", logger)
```
krasi-georgiev, line 48: No need to choose a differnt name and lets use the idea for `const ComponentName = "DataServer"`
```suggestion
	logger, err := util.ApplyFilter(*config, ComponentName, logger)
```

later in the code it can be like
```

logger:       log.With(filterLogger, "component", ComponentName)}, nil
```

pkg/db/db.go
```diff
@@ -52,13 +55,18 @@ func Open(file string) (DB, error) {
 		return nil, err
 	}
 
-	i := &impl{db: db, log: util.NewLogger("db", "DB")}
-	i.log.Info("Created DB at path: %s\n", file)
+	filterLogger, err := util.ApplyFilter(*config.GetConfig(), "dbDB", util.NewLogger())
```
krasi-georgiev, line 58: instead of this please pass the config and the logger to  to `func Open(lgr log.Logger,  cfg config.Config,file string)`, 
Also use ComponenName
```suggestion
	lgr, err := util.ApplyFilter(cfg, ComponenName, lgr)
```

pkg/db/localDataProxy.go
```diff
@@ -0,0 +1,78 @@
+// Copyright (c) The Tellor Authors.
+// Licensed under the MIT License.
```
krasi-georgiev, line 2: this file looks unintentional

krasi-georgiev (commented):
To avoid circular imports create a package logger and put `ApplyFilter` and `New` in it so we have 	`logger.ApplyFilter` and `logger.New`

 

the logger and config should be passed to the component constructors  use global constant for the ComponentName

So in tl-dr in `cmds.go` use `lgr:=util.NewLogger()` and `cfg:=config.GetConfig`

Then pass these when you initialize a component. for example on package ops
```
package ops
New(lgr,cfg){
	lgr = logger.ApplyFilter(cfg, ComponentName, lgr)
	lgr = log.With(lgr, "component", ComponentName)
}
```

pkg/apiOracle/valueOracle.go
```diff
@@ -76,18 +78,22 @@ func writeOutHistory() {
 	psrSavedDataTmp := psrSavedData + ".tmp"
 	err = ioutil.WriteFile(psrSavedDataTmp, data, 0644)
 	if err != nil {
-		logger.Error("failed to write out PSR values to %s: %s", psrSavedDataTmp, err.Error())
+		level.Error(logger).Log(
+			"msg", "failed to write out PSR values",
+			"psrSavedDataTmp", psrSavedDataTmp,
+			"err", err.Error(),
```
krasi-georgiev, line 84: Here and few other places

```suggestion
			"err", err,
```

pkg/apiOracle/valueOracle.go
```diff
@@ -76,18 +78,22 @@ func writeOutHistory() {
...
+		)
 		return
 	}
 	// Rename tmp file to old file (should be atomic on most modern OS)
 	err = os.Rename(psrSavedDataTmp, psrSavedData)
 	if err != nil {
-		logger.Error("failed move new PSR save onto old: %s", err.Error())
+		level.Error(logger).Log("msg", "failed move new PSR save onto old", "err", err.Error())
```
krasi-georgiev, line 91:
```suggestion
		level.Error(logger).Log("msg", "move new PSR save onto old", "err", err)
```

We just describe the action see the coding guidelines. No log msg should include the words fail, error, problem etc.

pkg/apiOracle/valueOracle.go
```diff
@@ -100,11 +106,15 @@ func EnsureValueOracle() error {
 		return nil
 	}
 
-	cfg := config.GetConfig()
+	filterLog, err := logging.ApplyFilter(*cfg, ComponentName, logger)
+	if err != nil {
+		return errors.Wrap(err, "creating filter logger")
+	}
+	logWith := log.With(filterLog, "component", "valueOracle")
```
krasi-georgiev, line 113: no need to create new variables. Here and few other places
```suggestion
	logger, err := logging.ApplyFilter(*cfg, ComponentName, logger)
	if err != nil {
		return errors.Wrap(err, "creating filter logger")
	}
	logger := log.With(filterLog, "component", "valueOracle")
```

pkg/db/proxy.go
```diff
@@ -69,27 +69,32 @@ type remoteImpl struct {
...
 // OpenRemoteDB establishes a proxy to a remote data server.
-func open(cfg *config.Config, localDB DB, isRemote bool) (DataServerProxy, error) {
-	rdbLog = util.NewLogger("db", "RemoteDBProxy")
+func open(logger log.Logger, cfg *config.Config, localDB DB, isRemote bool) (DataServerProxy, error) {
+	filterLog, err := logging.ApplyFilter(*cfg, ComponentName, logger)
+	if err != nil {
+		return nil, errors.Wrap(err, "applying filter to logger")
+	}
+
+	rdbLog := log.With(filterLog, "component", "RemoteDBProxy")
```
krasi-georgiev, line 93:
```suggestion
	rdbLog := log.With(filterLog, "component", ComponentName)
```
please make all component names constants.

pkg/rpc/mock.go
```diff
@@ -185,27 +190,27 @@ func (c *mockClient) CallContract(ctx context.Context, call ethereum.CallMsg, bl
...
 		return []byte{}, nil
 	}
 
 	switch fn {
 	case balanceAtFN:
 		{
 			return meth.Outputs.Pack(c.tokenBalance)
-			//mockClientLog.Debug("Getting balance from contract")
+			//level.Debug(c.logger).Log("msg", "getting balance from contract")
 			//return math.PaddedBigBytes(math.U256(c.tokenBalance), 32), nil
```
krasi-georgiev, line 202: any place with commented log of printf you can delete.

pkg/tracker/psrs.go
```diff
@@ -134,9 +135,7 @@ func TimeWeightedAvg(interval time.Duration, weightFn func(float64) (float64, fl
 		result.Volume = maxVolume
 		// if math.Min(weightSum/targetWeight, 1.0) < .5{
 		// 	values,_ := apiOracle.GetNearestTwoRequestValue(apis[0].Identifier, at)
-		// 	fmt.Println("not enough data for time series, series starts : ", values.Created)
 		// }
```
krasi-georgiev, line 138: can delete these lines as well.

pkg/tracker/psrs.go
```diff
@@ -179,14 +178,15 @@ func MedianAt(apis []*IndexTracker, at time.Time) (apiOracle.PriceInfo, float64)
 }
 
 func ManualEntry(apis []*IndexTracker, at time.Time) (apiOracle.PriceInfo, float64) {
+	logger := logging.NewLogger()
```
krasi-georgiev, line 181: Pass the logger as an argument.

krasi-georgiev (commented):
can you also update the configs/config.json

and also need to remove the logLevel flag from the cli flags.

pkg/util/logConfig.go
```diff
@@ -11,6 +11,7 @@ var (
 	sharedConfig *LogConfig
 )
 
+/*
```
krasi-georgiev, line 14: this whole file should be deleted now

pkg/util/logger.go
```diff
@@ -12,7 +12,6 @@ import (
 	"github.com/go-kit/kit/log"
```
krasi-georgiev, line 12: This whole file should be deleted. If anything is needed it should go in the pkg/logging

pkg/logging/logging.go
```diff
@@ -0,0 +1,41 @@
...
+	"os"
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/pkg/errors"
+	"github.com/tellor-io/telliot/pkg/config"
+)
+
+// NewLogger create a new logger
```
krasi-georgiev, line 16: Please follow Golang commentary guidelines.

Here and few other places
```suggestion
// NewLogger create a new logger.
```

pkg/ops/miningManager.go
```diff
@@ -77,12 +78,17 @@ func CreateMiningManager(
 	account *rpc.Account,
 ) (*MiningMgr, error) {
 
-	group, err := pow.SetupMiningGroup(cfg, exitCh)
+	powFilterLogger, err := logging.ApplyFilter(*cfg, ComponentName, logger)
```
krasi-georgiev, line 81: here and few other places
```suggestion
	logger, err := logging.ApplyFilter(*cfg, ComponentName, logger)
```

pkg/util/httpRetriever.go
```diff
@@ -45,24 +49,33 @@ func _recReq(req *HTTPFetchRequest, expiration time.Time) ([]byte, error) {
...
+		level.Warn(logger).Log(
+			"msg", "problem fetching data",
+			"queryURL", req.QueryURL,
+			"err", err,
+		)
 		now := time.Now()
 		if now.After(expiration) {
-			httpFetchLog.Error("Timeout expired, not retrying query and passing error up")
+			level.Error(logger).Log("msg", "timeout expired, not retrying query and passing error up")
 			return nil, err
```
krasi-georgiev, line 60: In places like this where we return the error there is no need to log it as the caller will log it anyway. Instead should just wrap it.
```suggestion
			return nil, errors.Wrap(err,"timeout expired, not retrying query")
```

pkg/ops/miningManager.go
TheRocketCat: problem with that is that we user logger twice, to make both powFilterLogger and opsFilterLogger. So redefining the variable will mess up the creation of opsFilterLogger.

krasi-georgiev: Aah I see the issue. I think we don't actually need filtering for the sub-component for now lets apply log filtering only for the `MiningMgr`

krasi-georgiev (approved):

PR #390, opened 2021-01-26
removed stratum code

JGcarv:


krasi-georgiev (commented):
I think the signature of GetWork can also change since now it doesn't need a channel.

krasi-georgiev (approved):

PR #396, opened 2021-02-03
Adding contracts infra for 2e2 test

JGcarv:


JGcarv:
I think we can close this as a lot will need to be done again. Starting with a fresher branch might be worth it

krasi-georgiev:
ok

PR #401, opened 2021-02-08
Add the contracts generating tool as a package

krasi-georgiev:
This will allow us to reuse it across projects.

Also added abi generation function so this can be used when interacting
with the contracts.


krasi-georgiev:
opened by mistake

PR #402, opened 2021-02-10
Renaming Contracts to be more explicit about master and curent

JGcarv:
Making this PR to sort of standardize the nomenclature between all repos. 

In TellorCore, the contract called `TellorMaster` is actually the one that we were previously calling `TellorProxy`  here. The one called `tellorMaster` here is what we call simply `Tellor.sol` in the core, which I renamed to `tellorCurrent` here, as it supposed to track the latest version on chain.

The whole mess was making me quite confused while adding e2e and I bet it'll be a source of confusion in the future too.


krasi-georgiev:
yep I agree.you also need to run `make generate` locally

krasi-georgiev (approved):

krasi-georgiev:
lets merge after https://github.com/tellor-io/telliot/pull/399 as this will cause merge conflicts and this PR will be easier to rebase and fix the conflict.

PR #416, opened 2021-02-24
Fix profit calculation to use more precision with floating points

### LINKED ISSUE #415: Incorrect profit margin calculation
The `profit()` function returns an incorrect `profitPercent` value. There are two issues:

- The profitPercent is computed using an integer divide, so the result is rounded to the nearest 100% - usually either 100%, 0, or 100%.  It should use a floating point divide.
- When there is an error, a `profitPercent` -1% is returned.  It should return -100%, which is the worst case.

lodainn:
Fixes #415


krasi-georgiev (approved):

krasi-georgiev:
Thanks!

PR #427, opened 2021-03-19
add transactor interface

hhio618:
### Description
- Add the transactor component


krasi-georgiev (approved):

krasi-georgiev:
For a bit more context we are adding this so that people can use the submitter, but still plug in their logic how to create and handle the actual blockchain transaction

PR #434, opened 2021-04-03
Add alertmanager, Telegram alerting

hhio618:
### Description
This will add:
- Prometheus Alertmanager
- A Telegram Alertmanager receiver using [this repo](https://github.com/metalmatze/alertmanager-bot)

### DIFF
--- a/configs/manifests/alerting-persist.yml
+++ b/configs/manifests/alerting-persist.yml
@@ -1,0 +1,10 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: alertmanager-bot
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
--- a/docs/setup-and-usage.md
+++ b/docs/setup-and-usage.md
@@ -105,6 +105,17 @@ kubectl apply -f configs/manifests/monitoring-persist.yml
 kubectl apply -f configs/manifests/monitoring.yml
 ```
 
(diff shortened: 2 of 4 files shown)


configs/manifests/alerting-persist.yml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
```
krasi-georgiev, line 2: What would we loose if we don't use persistent volume?

docs/setup-and-usage.md
```diff
@@ -105,6 +105,17 @@ kubectl apply -f configs/manifests/monitoring-persist.yml
 kubectl apply -f configs/manifests/monitoring.yml
 ```
 
+* Optionally deploy the alerting manager and get alerts on Telegram.
```
krasi-georgiev, line 108: Maybe also add a link to the bot readme for how to setup. IIRC there was some additional setup right?

configs/manifests/alerting-persist.yml
hhio618: It is needed for saving list of chat that the bot added to them. see [here.](https://github.com/metalmatze/alertmanager-bot/blob/11438e166eac991c6540e10dcfdcabcb2006e349/pkg/telegram/chats.go)

docs/setup-and-usage.md
hhio618: I added their repo url. hopefully no additional setup required for bot to be working.

configs/manifests/alerting-persist.yml
krasi-georgiev: ok lets make it really small as I don't think it will use much space. something like 100MB

themandalore (approved):

Issue #426, opened 2021-03-19
Profit calculation is incorrect because the gas used varies on even and odd oracle blocks.


krasi-georgiev:
for slot five the gas varies
1,671,890 for some oracle blocks 
and
1,776,264 for other oracle blocks.

It seems that we need to adjust the profit calc to save the gas used in a different way.

JGcarv:
Wait, seriously? I have no idea why that happens

krasi-georgiev:
yep , look at the gas usage for these 2 transactions

https://etherscan.io/tx/0x938a71c9c02a4a1fd561ff7f682e3a7f66e3b8efad04c3d7487f430e39f339be
https://etherscan.io/tx/0x8e5aeca09d9eb9e0679e2d68dbb1cfe5f61b8e2b1c637819843a8bcee1133be4

It is actually quite random. some use 100-110k more gas.

Do you think this is normal or some extra variable causing this?

krasi-georgiev:
aah could it be because the values for some data Ids have more numbers in them? like the TVL

PR #451, opened 2021-06-13
Add the gas estimation script

hhio618:
### Description
Add a script to call the geth gas estimation API by using random index data.
### Refs
#426


krasi-georgiev:
No I didn't mean to add the script but to actually use it 
https://github.com/tellor-io/telliot/blob/master/pkg/reward/reward.go#L73

after adding can remove
https://github.com/tellor-io/telliot/blob/master/pkg/reward/reward.go#L91

pkg/reward/reward.go
```diff
@@ -122,7 +216,7 @@ func (self *Reward) rewardInEth1e18() (*big.Int, error) {
 }
 
 func (s *Reward) Slot() (*big.Int, error) {
```
krasi-georgiev, line 218: don't think we need this function anymore right?

pkg/reward/reward.go
```diff
@@ -4,44 +4,66 @@
...
+	// Getting abi.
+	abi, _ := abi.JSON(strings.NewReader(tellor.TellorABI))
 	return &Reward{
 		aggr:           aggr,
 		logger:         log.With(logger, "component", ComponentName),
 		contractCaller: contractCaller,
 		gasUsed:        make(map[int64]*big.Int),
+		client:         client,
+		abi:            abi,
+		account:        account,
```
krasi-georgiev, line 50: I think instead of adding it here this should be added as an argument to `GasUsed`
`GasUsed(....,account)`

pkg/reward/reward.go
```diff
@@ -70,12 +92,84 @@ func (self *Reward) Current(slot *big.Int, gasPriceEth1e18 *big.Int) (int64, err
...
+		}
+		if i%2 == 0 {
+			fmt.Println("gasLimit(random)", gasUsed)
+		} else {
+			fmt.Println("gasLimit(maxInt)", gasUsed)
+		}
+		i++
+		if i == 5 {
+			return nil, ErrNoDataForSlot{slot: slot.String()}
+		}
```
krasi-georgiev, line 170: this should never happen right? so we don't need it.

pkg/reward/reward.go
```diff
@@ -70,12 +92,84 @@ func (self *Reward) Current(slot *big.Int, gasPriceEth1e18 *big.Int) (int64, err
...
+		vars struct {
+			Challenge  [32]byte
+			RequestIds [5]*big.Int
+			Difficutly *big.Int
+			Tip        *big.Int
+		}
+		err error
+	)
+	// Getting current challenge.
+	for {
```
krasi-georgiev, line 110: I think we don't need a loop here, should just return on any error.

pkg/reward/reward.go
```diff
@@ -70,12 +92,84 @@ func (self *Reward) Current(slot *big.Int, gasPriceEth1e18 *big.Int) (int64, err
...
+			level.Debug(self.logger).Log("msg", "calling GetNewCurrentVariables", "err", err)
+			<-ticker.C
+			continue
+		}
+		break
 	}
 
-	return nil, ErrNoDataForSlot{slot: slot.String()}
+	i := 0
+	for {
```
krasi-georgiev, line 126: The same here - we don't need to retry just return on any error.

pkg/reward/reward.go
```diff
@@ -70,12 +92,84 @@ func (self *Reward) Current(slot *big.Int, gasPriceEth1e18 *big.Int) (int64, err
...
+			<-ticker.C
+			continue
+		}
+		reqVals := [5]*big.Int{
+			big.NewInt(math.MaxInt64),
+			big.NewInt(math.MaxInt64),
+			big.NewInt(math.MaxInt64),
+			big.NewInt(math.MaxInt64),
+			big.NewInt(math.MaxInt64),
+		}
```
krasi-georgiev, line 144: Add a comment on why we are not using actual values and just the max value.

IIRC this is because the gas calculation didn't change much between using random and actual values right?

pkg/reward/reward.go
```diff
@@ -70,12 +92,84 @@ func (self *Reward) Current(slot *big.Int, gasPriceEth1e18 *big.Int) (int64, err
...
+
+		packed, err := self.abi.Pack("submitMiningSolution", "", vars.RequestIds, reqVals)
+		if err != nil {
+			level.Debug(self.logger).Log("msg", "packing submitMiningSolution args", "err", err)
+			return nil, errors.Wrap(err, "packing submitMiningSolution args")
```
krasi-georgiev, line 149:
```suggestion
			return nil, errors.Wrap(err, "packing submitMiningSolution args")
```
When an error is returned no need to print a log as the caller will do that.

pkg/reward/reward.go
```diff
@@ -70,12 +92,84 @@ func (self *Reward) Current(slot *big.Int, gasPriceEth1e18 *big.Int) (int64, err
...
+		err error
+	)
+	// Getting current challenge.
+	for {
+		select {
+		case <-ctx.Done():
+			return nil, errors.New("context cancelled")
+		default:
+		}
+		vars, err = self.contractCaller.GetNewCurrentVariables(nil)
```
krasi-georgiev, line 116: use a context here in the CallOpts, don't pass nil

krasi-georgiev (commented):
We don't need the `SaveGasUsed` function anymore

And don't forget to add metrics the submitter for estimated gas used and actual gas used to we can monitor the difference and try to improve

pkg/reward/reward.go
hhio618: Yeah, right!

pkg/reward/reward.go
hhio618: Exactly, that's the reason!

pkg/reward/reward.go
```diff
@@ -37,33 +44,34 @@ type ContractCaller interface {
...
+func New(logger log.Logger, aggr aggregator.IAggregator, contractCaller ContractCaller) *Reward {
 	return &Reward{
 		aggr:           aggr,
 		logger:         log.With(logger, "component", ComponentName),
 		contractCaller: contractCaller,
 		gasUsed:        make(map[int64]*big.Int),
-		client:         client,
-		abi:            abi,
-		account:        account,
+		estimatedGasCostValue: promauto.NewGaugeVec(prometheus.GaugeOpts{
```
krasi-georgiev, line 53: we also need to record the actual gas used so we can compare them. You can get it from the TX receipt.

hhio618: I added it to the submitter. the estimation happens before the actual submit, so maybe we need to record its value somewhere in the submitter. IDK how it's possible to add it to the reward code?

krasi-georgiev: yes add it in the submitter, there we have the estimation and then when the tx is confirmed we have the actual cost

pkg/reward/reward.go
```diff
@@ -70,7 +71,7 @@ type Reward struct {
 }
 
 // Current returns the profit in percents based on the current TRB price.
-func (self *Reward) Current(ctx context.Context, slot *big.Int, gasPriceEth1e18 *big.Int, client contracts.ETHClient, account *eth.Account) (int64, error) {
+func (self *Reward) Current(ctx context.Context, slot *big.Int, gasPriceEth1e18 *big.Int, client *ethclient.Client, account *eth.Account) (int64, error) {
```
krasi-georgiev, line 74: the client is not used, is this a mistake?

pkg/reward/reward.go
```diff
@@ -101,7 +102,7 @@ func (self *Reward) Current(ctx context.Context, slot *big.Int, gasPriceEth1e18
 }
 
 // GasUsed estimates the gas needed by the transaction.
-func (self *Reward) GasUsed(ctx context.Context, slot *big.Int, client contracts.ETHClient, account *eth.Account) (*big.Int, error) {
+func (self *Reward) GasUsed(ctx context.Context, slot *big.Int, client *ethclient.Client, account *eth.Account) (*big.Int, error) {
```
krasi-georgiev, line 105: here as well? why do we need it?

pkg/reward/reward.go
hhio618: I used it in line#123, but i can remove it as we have the gas price argument.

hhio618: Oh, actually I used it to call `EstimateGas` method!

pkg/cli/cli.go
```diff
@@ -737,6 +737,23 @@ func (self mineCmd) Run() error {
 				profitTracker.Stop()
 			})
 
+			// Reward tracker.
```
krasi-georgiev, line 740: also always run it in the `dataserver` command and in the `mine` command run only when not connected to a remote db as when connected to a remote DB the remote DB already runs it, look at my new PR to improve this https://github.com/tellor-io/telliot/pull/468

krasi-georgiev (commented):
Lets add a note to add a test for this. We would need to run a simulated backed -> deploy the tellot contract -> submit values for all 5 slots -> check that the tracker has recorded the correct values. Not going to do it in this PR, but at least create a file ..._test and add a note to add a test for this.

pkg/tracker/reward/reward.go
```diff
@@ -0,0 +1,426 @@
...
+	)
+	// Getting current challenge.
+	vars, err = self.contractInstance.GetNewCurrentVariables(&bind.CallOpts{Context: self.ctx})
+	if err != nil {
+		level.Error(self.logger).Log("msg", "call GetNewCurrentVariables", "err", err)
+	}
+	gasPrice, err := self.client.SuggestGasPrice(self.ctx)
+	if err != nil {
+		level.Error(self.logger).Log("msg", "call client.SuggestGasPrice", "err", err)
+	}
```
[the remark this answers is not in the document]
hhio618: I'm not sure about it. let's try with a hardcoded value

pkg/cli/cli.go
hhio618: Ok, also we need to query these values in the submitter, so we always need a reward instance in the submitter or we can move the query part to a separate package.

krasi-georgiev: yes I meant to have an instante in the dataserver command and another instance here

pkg/tracker/reward/reward.go
```diff
@@ -0,0 +1,418 @@
...
+			}
+			return
+		}
+		if errC := appender.Commit(); errC != nil {
+			err = errors.Wrap(err, "db append commit failed")
+		}
+	}()
+	lbls := labels.Labels{
+		labels.Label{Name: "__name__", Value: "gas_usage_estimated"},
+		labels.Label{Name: "__slot__", Value: slot.String()},
```
krasi-georgiev, line 311: only the name needs to have underscored, this is prometheus thing
Also rename it in the query
```suggestion
		labels.Label{Name: "slot", Value: slot.String()},
```

pkg/tracker/reward/reward.go
```diff
@@ -0,0 +1,418 @@
...
+		big.NewInt(math.MaxInt64),
+	}
+	packed, err := abi.Pack("submitMiningSolution", "", vars.RequestIds, reqVals)
+	if err != nil {
+		return errors.Wrap(err, "packing submitMiningSolution args")
+	}
+	data := ethereum.CallMsg{
+		From: self.addrs[0],
+		To:   &self.contractInstance.Address,
+		// Hardcoded gas price. looks like it doesn't matter in the calculation.
```
krasi-georgiev, line 275:
```suggestion
		// Hardcoded gas price. Looks like it doesn't matter in the calculation.
```

pkg/tracker/reward/reward.go
```diff
@@ -0,0 +1,418 @@
...
+			level.Info(logger).Log("msg", "re-subscribed to events")
+		case event := <-events:
+			err := self.recordGasUsage(event)
+			if err != nil {
+				level.Error(self.logger).Log("msg", "record gas usage", "err", err)
+			}
+			ctx, _ := context.WithTimeout(self.ctx, 2*time.Second)
+			err = self.recordGasUsageEstimated(ctx)
+			if err != nil {
+				level.Error(self.logger).Log("msg", "record gas usage estimated", "err", err)
```
krasi-georgiev, line 178:
```suggestion
				level.Error(self.logger).Log("msg", "record gas usage estimation", "err", err)
```

pkg/tracker/reward/reward.go
```diff
@@ -0,0 +1,418 @@
...
+	_, err = appender.Append(0, lbls, ts, float64(gasUsed))
+	if err != nil {
+		return errors.Wrap(err, "append values to the DB")
+	}
+	return nil
+}
+
+func (self *RewardTracker) recordGasUsageEstimated(ctx context.Context) error {
+	// Getting abi.
+	abi, _ := abi.JSON(strings.NewReader(tellor.TellorABI))
```
krasi-georgiev, line 236: unchecker error.

pkg/tracker/reward/reward.go
```diff
@@ -0,0 +1,418 @@
...
+				}
+				break
+			}
+			level.Info(logger).Log("msg", "re-subscribed to events")
+		case event := <-events:
+			err := self.recordGasUsage(event)
+			if err != nil {
+				level.Error(self.logger).Log("msg", "record gas usage", "err", err)
+			}
+			ctx, _ := context.WithTimeout(self.ctx, 2*time.Second)
```
krasi-georgiev, line 175:
```suggestion
	err = self.recordGasUsageEstimated(self.ctx)
	
	func ... recordGasUsageEstimated(ctx ){
		ctx, cncl := context.WithTimeout(ctx, 2*time.Second)
		defer cncl()
	}
```
here we should defer cncl() and for this need to move the context creatio inside the function

krasi-georgiev:
failing lint tests

pkg/cli/cli.go
krasi-georgiev: this should be in the `if cfg.Db.RemoteHost == "" {` block

pkg/tracker/reward/reward.go
```diff
@@ -0,0 +1,412 @@
...
+		tsDB:             tsDB,
+		engine:           engine,
+		aggr:             aggr,
+	}, nil
+}
+
+func (self *RewardTracker) Start() error {
+	level.Info(self.logger).Log("msg", "starting")
+
+	go self.monitorGas()
```
krasi-georgiev, line 108: don't think you need a new function here , all the logic can be just here.

pkg/tracker/reward/reward.go
```diff
@@ -0,0 +1,412 @@
...
+func (self *RewardTracker) recordGasUsage(event *tellor.TellorNonceSubmitted) error {
+	receipt, err := self.client.TransactionReceipt(self.ctx, event.Raw.TxHash)
+	if err != nil {
+		return errors.Wrap(err, "receipt retrieval")
+	} else if receipt != nil && receipt.Status == types.ReceiptStatusSuccessful { // Failed transactions cost is monitored in a different process.
+		// tx, _, err := self.client.TransactionByHash(self.ctx, event.Raw.TxHash)
+		// if err != nil {
+		// 	level.Error(self.logger).Log("msg", "get transaction by hash", "err", err)
+		// 	return
+		// }
```
krasi-georgiev, line 192:
```suggestion
```

pkg/tracker/reward/reward.go
```diff
@@ -0,0 +1,412 @@
...
+				}
+				break
+			}
+			level.Info(logger).Log("msg", "re-subscribed to events")
+		case event := <-events:
+			err := self.recordGasUsage(event)
+			if err != nil {
+				level.Error(self.logger).Log("msg", "record gas usage", "err", err)
+			}
+			err = self.recordGasUsageEstimated(self.ctx)
```
krasi-georgiev, line 175: don't need to pass context here as can get it directly in `recordGasUsageEstimated`
```suggestion
			err = self.recordGasUsageEstimated()
```

pkg/tracker/reward/reward.go
```diff
@@ -0,0 +1,412 @@
...
+	if err != nil {
+		return errors.Wrap(err, "append values to the DB")
+	}
+	return nil
+}
+
+func (self *RewardTracker) recordGasUsageEstimated(ctx context.Context) error {
+	ctx, cncl := context.WithTimeout(ctx, 2*time.Second)
+	defer cncl()
+	// Getting abi.
```
krasi-georgiev, line 236:
```suggestion
```
I think this and the other comments here are not necessary as the code is self explanatory

pkg/tracker/reward/reward.go
```diff
@@ -0,0 +1,412 @@
...
+		return errors.Wrap(err, "getting current slot")
+	}
+
+	// Getting current challenge.
+	vars, err := self.contractInstance.GetNewCurrentVariables(&bind.CallOpts{Context: ctx})
+	if err != nil {
+		return errors.Wrap(err, "call GetNewCurrentVariables")
+	}
+	// We don't have actual values here, so we'll use math.MaxInt64 as our values.
+	// As we tested, this won't make any significant difference in the estimated gas value!
```
krasi-georgiev, line 254: I prefer if we don't use `we`
```suggestion
	// Use arbitrary values as when tested didn't make a noticeable difference between using real vs arbitrary values.
```

pkg/tracker/reward/reward.go
```diff
@@ -0,0 +1,412 @@
...
+		level.Debug(self.logger).Log("msg", "adding gas used", "gasUsed", receipt.GasUsed)
+		if err := self.addGasUsed(event.Slot.String(), receipt.GasUsed); err != nil {
+			return errors.Wrap(err, "adding gas used")
+		}
+		return nil
+	}
+	return errors.New("transaction not yet mined")
+}
+
+func (self *RewardTracker) addGasUsed(slot string, gasUsed uint64) error {
```
krasi-georgiev, line 202: lets extract this as a function in the DB package like
```
DB.Add(ctx, logger, appender, lbls , value uint64) error {
	.....
}
``

this will allow removing some duplicated code here and in few other places that add to the db.

File: pkg/tracker/reward/reward.go [hunk 25]
```diff
@@ -0,0 +1,412 @@
...
+			}
+		}
+	}
+}
+
+func (self *RewardTracker) recordGasUsage(event *tellor.TellorNonceSubmitted) error {
+	receipt, err := self.client.TransactionReceipt(self.ctx, event.Raw.TxHash)
+	if err != nil {
+		return errors.Wrap(err, "receipt retrieval")
+	} else if receipt != nil && receipt.Status == types.ReceiptStatusSuccessful { // Failed transactions cost is monitored in a different process.
```
krasi-georgiev, line 187:
```suggestion
	} else if receipt != nil && receipt.Status == types.ReceiptStatusSuccessful {
```
please review all new added code for such unrelated comments and nits.

krasi-georgiev (commented):
remove the .env file

krasi-georgiev (commented):
change that comment

```
	// Dispute tracker.
			// Run it only when not connected to a remote DB.
```
to

```
// Run some component only when not connected to a remote DB.
```

pkg/db/db.go
```diff
@@ -50,3 +57,30 @@ func NewRemoteDB(cfg Config) (storage.SampleAndChunkQueryable, error) {
 		func() (i int64, err error) { return 0, nil },
 	), nil
 }
+
+func Add(ctx context.Context, logger log.Logger, tsdb *tsdb.DB, lbls labels.Labels, value float64) error {
```
krasi-georgiev, line 61: can you also use this function in  recordInterval, recordValue

pkg/cli/mine.go
```diff
@@ -230,6 +243,15 @@ func (self mineCmd) Run() error {
 
 				psr := psrTellor.New(loggerWithAddr, cfg.PsrTellor, aggregator)
 
+				_tsDB, ok := tsDB.(*tsdb.DB)
+				if !ok {
+					return errors.New("tsdb is not a writable DB instance")
+				}
```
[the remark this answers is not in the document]
hhio618: Changed

pkg/tracker/reward/reward.go
```diff
@@ -0,0 +1,358 @@
...
+
+type Config struct {
+	LogLevel string
+}
+
+type RewardTracker struct {
+	client           *ethclient.Client
+	logger           log.Logger
+	contractInstance *contracts.ITellor
+	abi              abi.ABI
```
krasi-georgiev, line 48:
```suggestion
```
unused

pkg/tracker/reward/querier.go
```diff
@@ -0,0 +1,170 @@
...
+
+func (e ErrNoDataForSlot) Error() string {
+	return "no data for gas used for slot:" + e.slot
+}
+
+// GasUsed estimates the gas needed by the transaction.
+func (self *RewardQuerier) GasUsed(ctx context.Context, slot *big.Int) (*big.Int, error) {
+	query, err := self.engine.NewInstantQuery(
+		self.tsDB,
+		`last_over_time(gas_usage_estimated{slot="`+slot.String()+`"}[1d]) - `+`last_over_time(gas_usage_actual{slot="`+slot.String()+`"}[1d])`,
```
krasi-georgiev, line 121: Why is this like this?
This query will return the refund no? we need gas_usage_estimated - the refund right?

pkg/tracker/reward/reward.go
```diff
@@ -0,0 +1,358 @@
...
+		"gasPrice", gasPriceEth1e18,
+		"profit", fmt.Sprintf("%.2e", float64(profit.Int64())),
+		"profitMargin", profitPercent,
+	)
+
+	return profitPercent, nil
+}
+
+// GasUsed estimates the gas needed by the transaction.
+func (self *RewardTracker) GasUsed(ctx context.Context, slot *big.Int) (*big.Int, error) {
```
krasi-georgiev, line 306: this and few other functions are unused so can remove them.

krasi-georgiev:
wondering if we can come up with some good test for the reward calculation?

themandalore (approved):

krasi-georgiev:
this Pr is outdated and not working, I tested it locally.

PR #459, opened 2021-06-18
fix broken links

hhio618:
### Description
This should fix broken TellorDocs link [here](https://docs.tellor.io/tellor/telliot/internal-architecture#trackers) and [here](https://docs.tellor.io/tellor/telliot/contributing#project-development-go-tools).


docs/SUMMARY.md
```diff
@@ -4,8 +4,8 @@
 
 ## Docs
 
-* [Internal architecture](internal-architecture.md)
-  * [Index Tracker](index-tracker.md)
+* [Internal architecture](internal-architecture/README.md)
+  * [Index Tracker](internal-architecture/index-tracker.md)
```
krasi-georgiev, line 8: hhm, shoultn't this happen in the sync script that needs to fix these as they get broken when the docs get added to the main ones right?

docs/contributing.md
```diff
@@ -101,5 +101,6 @@ git commit
 
 ### Project development go tools
 
-The project uses [Bingo](.bingo/README.md) for adding any go tools required by the project.
+The project uses [Bingo](https://github.com/tellor-io/telliot/blob/master/.bingo/README.md) for adding any go tools required by the project.
```
krasi-georgiev, line 104: can we keep it relative here and just replace any relative links with absolute paths in the sync with the main docs?

docs/SUMMARY.md
hhio618: Looks like this is the default gitbook way of adding internal pages (you can see it by yourself when adding an internal page using gitbook dashboard). and also this will add complication to the sync process.

docs/contributing.md
hhio618: There are page links that need to stay relative. this could result in broken links in the TellorDocs repo!

docs/SUMMARY.md
krasi-georgiev: so how come it works even without this extra folder before adding it to the main docs?
https://docs.tellor.io/docs-preview/v/docs-preview/

docs/contributing.md
krasi-georgiev: all right I see what you mean so yeah lets leave it like this then.

docs/SUMMARY.md
hhio618: I'm not sure why this works here and not there, maybe we should ask from the Gitbook discord.

krasi-georgiev: yep , please do and update the issue with the answer

PR #468, opened 2021-06-25
don't run a dispute tracker when connected to a remote DB

krasi-georgiev:
The remote DB already records these so no need to do it locally as well.


krasi-georgiev:
added the changes in https://github.com/tellor-io/telliot/pull/466

PR #479, opened 2021-08-02
Update manualData.json

Spuddyminer:


owenburton (changes requested):
Change the VALUE to have a recurring three: `114.75833333333333`

PR #489, opened 2021-08-20
Updated AMPL_USD_VWAP Feed to Pull VWAP of last full day

imaginebeingatcomputers:
This will ensure the AMPL_USD_VWAP feed displays the VWAP of the last full UTC day.

### DIFF
--- a/configs/index.json
+++ b/configs/index.json
@@ -104,8 +104,12 @@
                 "param": "$.content[0][vwap]"
             },
             {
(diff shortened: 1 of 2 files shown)


themandalore (approved):

imaginebeingatcomputers:
I've also added the Crypto Compare VWAP API to this PR. During my investigation into Ampleforth's price timing I noticed they also referenced this API in their documentation. The median result of the 3 APIs come around 1 tenth of a cent closer to the the Ampleforth official oracle on average.

PR #496, opened 2021-09-01
Calculating time vars per request for anyblock, and cryptocompare

imaginebeingatcomputers:
This removes the ExpandTimeVars functions, and instead calculates time vars conditionally for anyblock and cryptocompare. Timestamp variables are then added to the request instead of being expanded via env variables.

### DIFF
--- a/configs/index.json
+++ b/configs/index.json
@@ -104,11 +104,11 @@
                 "param": "$.content[0][vwap]"
             },
             {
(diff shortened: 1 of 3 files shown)


krasi-georgiev:
This calculation is not right according to the AMPL documentation and it returns different results than their API, the docs clearly states, TWAP is calculated at the time of the request and not EOD or BOD, I messaged them in their channel , but didn't get a reply. jus now almost all submitted values are again off by 2-3%

PR #497, opened 2021-09-17
remove cryptocompare api

oraclown:


krasi-georgiev:
if you use the toTS url argument it fixes the issue

https://min-api.cryptocompare.com/data/dayAvg?fsym=AMPL&tsym=USD&toTs=$BOD

themandalore:
> if you use the toTS url argument it fixes the issue
> 
> https://min-api.cryptocompare.com/data/dayAvg?fsym=AMPL&tsym=USD&toTs=$BOD

It was actually a request form the ampl team.  Apparently this API includes some exchanges they didn't like, so we're pulling it on their request (nothing wrong w/it on our end)

themandalore (approved):

